Issue #6, opened 2019-07-31
`request` should be refactored to use `parse_headers`


siddacious:
When this lib was split out from `Adafruit_CircuitPython_ESP32SPI` `parse_headers` was inlined into the `request` method. Since it was added back for external use, it should be used here as well.

siddacious:
(sorry @brentru I should have mentioned this in the review for `parse_headers`. I can take care of this later)

brentru:
@siddacious 
Could you link a previous version of how it was inlined into the `request` method? Seems its a standalone method, this is the last commit of it before it was removed:
https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/8dbeef02125e4f27d49fe468d78912afc366a200/adafruit_esp32spi/adafruit_esp32spi_requests.py 

Do you think it's useful externally too, or only within `request`?

siddacious:
Here is the section in question:
https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/f77d71377125a023d5cecdc72d541cd8ba97b104/adafruit_requests.py#L227

It's used externally in the wsgi server code in the Esp32 lib so yes it's useful externally? They're using them for different sides of the server/client divide (parsing headers for an incoming request vs. parsing the response to an outgoing request) but there is some obvious overlap there.

brentru:
Merged into master and released, closing...

PR #7, opened 2019-08-01
Refactor request method to use parse_headers

brentru:
Addressing https://github.com/adafruit/Adafruit_CircuitPython_Requests/issues/6

Request now uses `parse_headers`. 

Tested with `WSGIServer.py` on Adafruit CircuitPython 4.0.0 on 2019-05-20; Adafruit PyPortal with samd51j20.
https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/examples/server/esp32spi_wsgiserver.py

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -224,7 +224,7 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
         reason = ""
         if len(line) > 2:
             reason = line[2].rstrip()
-        self.parse_headers(sock)
+        parse_headers(sock)
 
         if line.startswith(b"Transfer-Encoding:"):
             if b"chunked" in line:
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -224,12 +224,11 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
         reason = ""
         if len(line) > 2:
             reason = line[2].rstrip()
-        parse_headers(sock)
-
-        if line.startswith(b"Transfer-Encoding:"):
-            if b"chunked" in line:
-                raise ValueError("Unsupported " + line)
-        elif line.startswith(b"Location:") and not 200 <= status <= 299:
+        resp.headers = parse_headers(sock)
+        if resp.headers.get("transfer-encoding"):
+            if "chunked" in resp.headers.get("transfer-encoding"):
+                raise ValueError("Unsupported " + str(line[0]))
+        elif resp.headers.get("location") and not 200 <= status <= 299:
             raise NotImplementedError("Redirects not yet supported")
 
     except:
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -227,7 +227,7 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
         resp.headers = parse_headers(sock)
         if resp.headers.get("transfer-encoding"):
             if "chunked" in resp.headers.get("transfer-encoding"):
-                raise ValueError("Unsupported " + str(line[0]))
+                raise ValueError("Unsupported " + resp.headers.get("transfer-encoding"))
         elif resp.headers.get("location") and not 200 <= status <= 299:
             raise NotImplementedError("Redirects not yet supported")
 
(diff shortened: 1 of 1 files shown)


adafruit_requests.py
```diff
@@ -224,23 +224,13 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
...
-                resp.headers[title] = content
-
-            if line.startswith(b"Transfer-Encoding:"):
-                if b"chunked" in line:
-                    raise ValueError("Unsupported " + line)
-            elif line.startswith(b"Location:") and not 200 <= status <= 299:
-                raise NotImplementedError("Redirects not yet supported")
+        parse_headers(sock)
+
+        if line.startswith(b"Transfer-Encoding:"):
```
mscosti: parse_headers will have already read through all of the lines from the header section of the request, and put them into the headers dict. 

In order to raise these errors for unsupported headers, you'll want to check their presence inside of the that dict.

something like
```
if "chunked" in resp.headers.get("transfer-encoding"):
     raise ValueError("Unsupported " + line)
elif resp.headers.get("location") and not 200 <= status <= 299:
     raise NotImplementedError("Redirects not yet supported")
```

adafruit_requests.py
```diff
@@ -224,23 +224,13 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
...
-                title = str(title.lower(), "utf-8")
-                content = str(content, "utf-8")
-                resp.headers[title] = content
-
-            if line.startswith(b"Transfer-Encoding:"):
-                if b"chunked" in line:
-                    raise ValueError("Unsupported " + line)
-            elif line.startswith(b"Location:") and not 200 <= status <= 299:
-                raise NotImplementedError("Redirects not yet supported")
+        parse_headers(sock)
```
mscosti: parse_headers returns a dict of the headers. We need `resp.headers` to be set to that dict
```
resp.headers = parse_headers(sock)
```

brentru:
@mscosti  Thanks for the review, made requested changes (modified them a bit to work properly).

-------------------------------------------------------
Test transfer-encoding
```
Fetching text from https://http-evader.semantic-gap.de/-BNBbebyzMwzSuz4dk7UyGt2idQ3EpHQa1KU1Etm0dzHohAtWjf5p
  File "code.py", line 39, in <module>
  File "/lib/adafruit_requests.py", line 270, in get
  File "/lib/adafruit_requests.py", line 236, in request
  File "/lib/adafruit_requests.py", line 230, in request
ValueError: Unsupported b'HTTP/1.2'
```

Test HTTP Transfer Requests:
```
Fetching text from https://jigsaw.w3.org/HTTP/300/301.html
  File "code.py", line 39, in <module>
  File "/lib/adafruit_requests.py", line 270, in get
  File "/lib/adafruit_requests.py", line 236, in request
  File "/lib/adafruit_requests.py", line 232, in request
NotImplementedError: Redirects not yet supported
```

adafruit_requests.py
```diff
@@ -224,23 +224,12 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
...
-
-            if line.startswith(b"Transfer-Encoding:"):
-                if b"chunked" in line:
-                    raise ValueError("Unsupported " + line)
-            elif line.startswith(b"Location:") and not 200 <= status <= 299:
-                raise NotImplementedError("Redirects not yet supported")
+        resp.headers = parse_headers(sock)
+        if resp.headers.get("transfer-encoding"):
+            if "chunked" in resp.headers.get("transfer-encoding"):
+                raise ValueError("Unsupported " + str(line[0]))
```
mscosti: oh, missed this one on my first pass. I think this is supposed to be the transfer encoding header value in the exception string.

brentru:
@mscosti  Moved the header value into the exception string...
```
Fetching text from https://http-evader.semantic-gap.de/-BNBbebyzMwzSuz4dk7UyGt2idQ3EpHQa1KU1Etm0dzHohAtWjf5p
  File "code.py", line 39, in <module>
  File "/lib/adafruit_requests.py", line 270, in get
  File "/lib/adafruit_requests.py", line 236, in request
  File "/lib/adafruit_requests.py", line 230, in request
ValueError: Unsupported chunked
```

PR #17, opened 2020-01-25
fix parsing error on lines like `b'X-Geo-Block-List:'` as received by github

ladyada:
We now pre-fill the contents to an empty string


ladyada:
yeah there's nothing after, and so split() gives an array of length 1

Issue #20, opened 2020-03-02
TypeError: wrong number of arguments when posting bytes array


jimbobbennett:
I have a `bytearray` containing image data from the [TTL camera sample](https://github.com/adafruit/Adafruit_CircuitPython_VC0706/blob/master/examples/vc0706_snapshot_filesystem.py). I want to POST this as an octet stream to a REST endpoint.

My code:

```python
import adafruit_requests as requests

...

buffer = bytearray()

headers={
            'Content-Type': 'application/octet-stream',
            'Prediction-Key': prediction_key
        }

r = requests.request('POST', prediction_endpoint, data=buffer, headers=headers)
```

When I run this I get the following:

```sh
......Traceback (most recent call last):
  File "code.py", line 85, in <module>
  File "adafruit_requests.py", line 242, in request
  File "adafruit_requests.py", line 224, in request
TypeError: wrong number of arguments
```

Running the code locally I've tracked down the error to this line:

```python
if data:
   sock.send(bytes(data, "utf-8"))
```

If I change the call to `bytes()` to remove the encoding parameter, it works.

```python
if data:
   sock.send(bytes(data))
```

Does the CircuitPython implementation of `bytes` have an encoding parameter?

jimbobbennett:
PR #21 raised to remove this parameter, if this is the correct fix

FoamyGuy:
Odd, I am able to run this at REPL and don't get the same error.
```
>>> bytes("hello", "utf-8")
b'hello'
```

jimbobbennett:
@FoamyGuy - I think it is because it is a `bytesarray`.

Just tested in the REPL:

```python
>>> b = bytearray(10)
>>> bytes(b, 'utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: wrong number of arguments 
```

FoamyGuy:
Looks like possible the issue is more related to the bytearray than the encoding param I use a bytebuffer instead of a string it reports the error:
```
>>> buffer = bytearray()
>>> bytes(buffer, "utf-8")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: wrong number of arguments
```

FoamyGuy:
This does not appear to be possible in CPython 3.7 either though the error is different than CircuitPython: 
```
[...]

PR #21, opened 2020-03-02
Removed encoding parameter from bytes call

jimbobbennett:
As dicsussed in #20, the encoding parameter passed to the `byes` call gives a **TypeError: wrong number of arguments**

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -221,7 +221,10 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
             sock.send(b"Content-Length: %d\r\n" % len(data))
         sock.send(b"\r\n")
         if data:
-            sock.send(bytes(data))
+            if isinstance(data, bytearray):
+                sock.send(bytes(data))
+            else:
+                sock.send(bytes(data, "utf-8"))
 
         line = sock.readline()
         # print(line)
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -221,7 +221,7 @@ def request(method, url, data=None, json=None, headers=None, stream=False, timeo
             sock.send(b"Content-Length: %d\r\n" % len(data))
         sock.send(b"\r\n")
         if data:
-            sock.send(bytes(data, "utf-8"))
+            sock.send(bytes(data))
 
         line = sock.readline()
         # print(line)


FoamyGuy:
Do you know of public API that accepts bytearray data via POST like this that you can share? If so I can test this out with a pyportal.

jimbobbennett:
I've been testing with the Azure Custom Vision service at [customvision.ai](https://customvision.ai). I created a model and posted to the prediction service. Source for how to do it is here: https://github.com/jimbobbennett/mandmcounter

This code has my fixed version of adafruit_requests in requests.py. You'd need to create an object detection model in Custom Vision and set the values in the secrets.py file for your WiFi and keys for the model.

I'm on a plane at the moment, but tomorrow I can provide detailed instructions on how to do it if that would help.

FoamyGuy:
Thank you! I think I should be able to get it tested using what you've posted. I will give it a try either tonight or tomorrow.

tannewt:
I'm ok removing this but would love to see an issue for adding support for the encoding parameter to CircuitPython. In general, we should ensure CircuitPython matches CPython.

jimbobbennett:
@tannewt - if you look at #20, @FoamyGuy has compared the behavior to CPython and CircuitPython, and the behavior is roughly the same.

If you call `bytes()` against a `string`, you can pass the encoding parameter and it works correctly on both platforms.
If you call `bytes()` against a `bytearray`, you get an error passing the encoding parameter, albeit a different, more descriptive error on CPython

The difference here is the error given by CircuitPython is different, and less instructive than the CPython one. Where is the right place to raise an issue to get the error message improved? Is it in https://github.com/adafruit/circuitpython?

FoamyGuy:
@tannewt do you mean adding `encoding` parameter to the `request()` method in this library? `bytes()` seems to be supporting the `encoding` parameter already, it just doesn't work with a bytearray as the data, that fails in CPython as well, but with a slightly different error.

tannewt:
> @tannewt - if you look at #20, @FoamyGuy has compared the behavior to CPython and CircuitPython, and the behavior is roughly the same.
> 
> If you call `bytes()` against a `string`, you can pass the encoding parameter and it works correctly on both platforms.
> If you call `bytes()` against a `bytearray`, you get an error passing the encoding parameter, albeit a different, more descriptive error on CPython
> 
> The difference here is the error given by CircuitPython is different, and less instructive than the CPython one. Where is the right place to raise an issue to get the error message improved? Is it in https://github.com/adafruit/circuitpython?

Ah! Ok. I misunderstood that. Sounds like it is close enough. It would be nice to ensure it is the same exception thrown but it's much less important.

This change looks good to me.

FoamyGuy:
Thank you @jimbobbennett!

PR #43, opened 2020-10-29
Fixed the Expected 01 but got 00 error

makermelissa:
This fixes the `Expected 01 but got 00` issue and increases the reliability of the library.


ladyada (approved):
didnt test, looks good

tannewt:
This doesn't work with other socket implementations because CPython sockets don't return a boolean: https://docs.python.org/3/library/socket.html#socket.socket.connect

PR #44, opened 2020-10-29
Toss out chunked data if we never read anything

ladyada:
we dont know how much is remaining if data is chunked, so avoid comparing `None` received


makermelissa (approved):
Looks good to me

tannewt:
Just saw this. I have the same fix in another PR with unit tests. Will merge this and leave the other PR for the tests. Thanks!

PR #46, opened 2020-10-31
Add, test support for chunk extensions

jepler:
The chunk line is read up to the terminating "\r\n", then any extensions are ignored by only using the string up to the first ";" (if any) when determining the length of the chunk.

Resources:
https://tools.ietf.org/html/rfc7230#section-4.1.1
https://www.boost.org/doc/libs/1_74_0/libs/beast/doc/html/beast/using_http/chunked_encoding.html

Thanks to @anecdata for asking on Discord why ";" was getting special treatment, leading to this investigation.  We don't know of a specific URL or service that is affected by this bug.

### DIFF
--- a/tests/chunk_test.py
+++ b/tests/chunk_test.py
@@ -57,14 +57,16 @@ def do_test_get_text(extra=b""):
     )
     assert r.text == str(text, "utf-8")
 
+
 def test_get_text():
     do_test_get_text()
 
+
 def test_get_text_extra():
-    do_test_get_text(b';blahblah; blah')
+    do_test_get_text(b";blahblah; blah")
 
 
-def do_test_close_flush(extra=b''):
+def do_test_close_flush(extra=b""):
     """Test that a chunked response can be closed even when the request contents were not accessed."""
     pool = mocket.MocketPool()
     pool.getaddrinfo.return_value = ((None, None, None, None, (ip, 80)),)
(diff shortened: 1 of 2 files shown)


tests/chunk_test.py
```diff
@@ -28,56 +32,58 @@ def _chunk(response, split):
 
 
 def test_get_text():
-    pool = mocket.MocketPool()
-    pool.getaddrinfo.return_value = ((None, None, None, None, (ip, 80)),)
-    c = _chunk(text, 33)
-    print(c)
-    sock = mocket.Mocket(headers + c)
-    pool.socket.return_value = sock
+    for extra in (b"", b";blahblah; blah"):
```
tannewt, line 35: Mind copying and pasting this into a new test function? That will lead to a clearer failure if/when it fails. Right now it's lumped into the basic case.

tannewt (changes requested):
Thanks for finding this issue!

tannewt:
@jepler Are you going to follow up on this?

tests/chunk_test.py
jepler: I've updated the test, let me know if it's still not what you're looking for.

tannewt (approved):
Works for me! Black is unhappy though.

tannewt (approved):
Thank you!

PR #59, opened 2021-01-08
Update requests_https_cpython.py

jfabernathy:
found some missing calls to fix he socket

### DIFF
--- a/examples/requests_https_cpython.py
+++ b/examples/requests_https_cpython.py
@@ -1,21 +1,35 @@
 # adafruit_requests usage with a CPython socket
-import socket
+import socketpool
 import ssl
 import adafruit_requests as requests
+import wifi
+import secrets
 
+# Get wifi details and more from a secrets.py file
+try:
+    from secrets import secrets
+except ImportError:
+    print("WiFi secrets are kept in secrets.py, please add them there!")
+    raise
+
+print("Connecting to %s"%secrets["ssid"])
+wifi.radio.connect(secrets["ssid"], secrets["password"])
+print("Connected to %s!"%secrets["ssid"])
+print("My IP address is", wifi.radio.ipv4_address)
+
+socket = socketpool.SocketPool(wifi.radio)
 https = requests.Session(socket, ssl.create_default_context())
 
 TEXT_URL = "https://httpbin.org/get"
 JSON_GET_URL = "https://httpbin.org/get"
 JSON_POST_URL = "https://httpbin.org/post"
 
-# print("Fetching text from %s" % TEXT_URL)
-# response = requests.get(TEXT_URL)
-# print("-" * 40)
-
-# print("Text Response: ", response.text)
-# print("-" * 40)
-# response.close()
+print("Fetching text from %s" % TEXT_URL)
+response = https.get(TEXT_URL)
+print("-" * 40)
+print("Text Response: ", response.text)
+print("-" * 40)
+response.close()
 
 print("Fetching JSON data from %s" % JSON_GET_URL)
 response = https.get(JSON_GET_URL)


examples/requests_https_cpython.py
```diff
@@ -1,21 +1,35 @@
 # adafruit_requests usage with a CPython socket
```
brentru, line 1: Change to `#adafruit_requests usage with SocketPool`

examples/requests_https_cpython.py
```diff
@@ -1,21 +1,35 @@
...
+print("My IP address is", wifi.radio.ipv4_address)
+
+socket = socketpool.SocketPool(wifi.radio)
 https = requests.Session(socket, ssl.create_default_context())
 
 TEXT_URL = "https://httpbin.org/get"
 JSON_GET_URL = "https://httpbin.org/get"
 JSON_POST_URL = "https://httpbin.org/post"
 
-# print("Fetching text from %s" % TEXT_URL)
```
brentru, line 12: Why was this deleted?

brentru (changes requested):
Instead of modifying the CPython (desktop python) example for S2 compatibility, could you make a new example in `examples` named `requests_https_wifi.py`

examples/requests_https_cpython.py
jfabernathy: no idea, if you're directing the comment to me

examples/requests_https_cpython.py
jfabernathy: it is impossible for me to tell when an example is for Python on something like a Raspberry pi or whether it's for the Metro ESP32-S2 or the older ESP32 boards.  I just took an example in that directory and found one without the esp32spi library and messed with it until it worked.  I would love to see the structure of these examples broken down into what works on what boards.  I'm just struggling to figure this out.

examples/requests_https_cpython.py
jfabernathy: I maybe thought it was a debug print and commented it out by mistake

jfabernathy (commented):
uncomment??

examples/requests_https_cpython.py
brentru: Not a debug, part of the example!

examples/requests_https_cpython.py
brentru: I understand, and that's why we're making this PR, to help differentiate between examples.

CPython refers to Python on a Raspberry Pi or Computer
CircuitPython refers to Python on a Microcontroller like an ESP32-S2.

This example would be named something like `requests_https_socketpool` since it uses the native socket pool  API (https://circuitpython.readthedocs.io/en/latest/shared-bindings/socketpool/index.html#socketpool.SocketPool.socket) for ESP32S2.

examples/requests_https_cpython.py
jfabernathy: understand

examples/requests_https_cpython.py
brentru: I think we should use the following suffixes for these examples: 
`_cpython` - Python on Computer/Raspberry Pi
`_esp32spi` - Using ESP32SPI as a Co-Processor
`_socketpool` - Using native CircuitPython sockets

Do you agree with this? Any tips on how I could clarify this more?

jfabernathy: I did not grasp the many types of pythons that there were.  My example was purely for use on the Metro ESP32-S2 that I have.  I have used some the Cpython on my Raspbeerry Pi.  I will defer to your best judgement on this but comments clarifying all this is good.

tannewt:
@jfabernathy Want to move this into a separate example still?

jfabernathy:
I'm not sure what you are asking.  I'm really confused about git hub and the whole process that it includes except for cloning repositories .  I was asked to create a pull request and I did having no idea what that meant or entailed. I think, if memory serves, I posted this example on the Adafruit Forum after I got help over there.  I've been confused ever since.

tannewt:
@jfabernathy In this PR you've edited the requests_https_cpython.py file so it works on CircuitPython. That is a useful example to have but we also want the example for CPython (normal desktop python). So, the edited version should be added as a new file to leave the cpython one as-is.

For GitHub background I recommend this guide: https://learn.adafruit.com/contribute-to-circuitpython-with-git-and-github

jfabernathy:
So how do we delete this pull request so the original file is undisturbed?  I'll work on a new example when I can.  I still don't know how I ended up on a Desktop example instead of a circuitpython example

jfabernathy:
on my copy of my example I've changed the comment from Cpython to circuitpython and I'm studying the guide on how to post an example to github.  I'll try to do that soon, but figuring out how github works is harder than writing Intel 386 assembler.

tannewt:
I'll close this. I think you should be able to do "Add file" in the top right-ish of this page: https://github.com/adafruit/Adafruit_CircuitPython_Requests/tree/master/examples

Issue #73, opened 2021-02-16
Out of sockets on 3rd Request


jfurcean:
The [MagTag Google Calendar Event Display Learn Guide](https://learn.adafruit.com/magtag-google-calendar-event-display/getting-the-date-time) is crashing with the following trace using CircuitPython 6.2.0-beta.2 and the latest libraries.

```
"Traceback (most recent call last):
  File "code.py", line 252, in <module>
  File "code.py", line 129, in get_calendar_events
  File "adafruit_requests.py", line 598, in get
  File "adafruit_requests.py", line 559, in request
  File "adafruit_requests.py", line 436, in _get_socket
  File "adafruit_requests.py", line 433, in _get_socket
RuntimeError: Out of sockets"
```

jfurcean:
I did a little bit of troubleshooting and I think the oauth request socket is being left open. I am not sure if this is a `adafruit_requests` issue or an `adafruit_oauth2`.

jfurcean:
This looks like a failure to close sockets. The below example crashes on the third request.

```python
from adafruit_magtag.magtag import MagTag

magtag = MagTag()
magtag.network.connect()

requests = magtag.network.requests

urls = ["https://wired.com/", "https://www.test.com/","https://apple.com/"]

for url in urls:
    print(url)
    resp = requests.get(url)

    resp.close()
```

```
Traceback (most recent call last):
  File "code.py", line 12, in <module>
  File "adafruit_requests.py", line 616, in get
  File "adafruit_requests.py", line 575, in request
  File "adafruit_requests.py", line 445, in _get_socket
  File "adafruit_requests.py", line 441, in _get_socket
RuntimeError: Out of sockets
```

anecdata:
Just for completeness, it's unrelated to MagTag. This also gets `Out of sockets`:
```python
import wifi
import socketpool
import ssl
import espidf
import adafruit_requests
from secrets import secrets

wifi.radio.connect(secrets['ssid'], secrets['password'])
pool = socketpool.SocketPool(wifi.radio)
requests = adafruit_requests.Session(pool, ssl.create_default_context())

urls = ["https://wired.com/", "https://www.test.com/", "https://apple.com/"]

for url in urls:
    print(url)
    resp = requests.get(url)
    resp.close()
```

jfurcean:
This should be fixed by PR #74.

PR #74, opened 2021-02-17
fix out of sockets issue

jfurcean:
Fixes problem when there are free sockets but are not closed when needed. See #73

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -405,6 +405,7 @@ class Session:
             self._close_socket(sock)
 
     def _get_socket(self, host, port, proto, *, timeout=1):
+        # pylint: disable=too-many-branches
         key = (host, port, proto)
         if key in self._open_sockets:
             sock = self._open_sockets[key]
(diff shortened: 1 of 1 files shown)


brentru:
@jfurcean  Could you clarify why you are `except`'ing a `RuntimeError: Out of sockets"`? Is this being thrown erroneously by CircuitPython?

Neradoc:
The assumption is that it is being thrown correctly, so it triggers a retry on the loop, which in turn causes a call to `self._free_sockets()` which would close the free sockets. Requests otherwise doesn't automatically close the sockets it uses, it keeps them and reuses them when you access the same host/port, leading to inevitably running out of sockets.

In 6.2 beta 1 the same code would cause a MemoryError on `sock.connect()` in that same loop which in turns would also cause the freeing of sockets. That change to a more specific error message is a consequence of the "Separate SSLSocket from Socket" PR.

brentru:
>  In 6.2 beta 1 the same code would cause a MemoryError on sock.connect() in that same loop which in turn would also cause the freeing of sockets. That change to a more specific error message is a consequence of the "Separate SSLSocket from Socket" PR.

I see, thank you for the clarification! I'm going to implement a similar patch within MiniMQTT.

brentru (approved):

PR #76, opened 2021-03-25
Fixes #48

### LINKED ISSUE #48: iface is not set correctly when using some examples
I'm not sure whether this is an error in the examples or in the revised legacy API.

`examples/requests_simpletest.py` and `examples/requests_advanced.py` both contain code:
```python
# Initialize a requests object with a socket and esp32spi interface
socket.set_interface(esp)
requests.set_socket(socket)
```
However, this code stopped working at some point during the revision of the legacy API. If I try the `simpletest`, I get
```
  File "requests_simpletest.py", line 51, in <module>
  File "adafruit_requests.py", line 636, in get
  File "adafruit_requests.py", line 535, in request
  File "adafruit_requests.py", line 420, in _get_socket
  File "adafruit_requests.py", line 604, in wrap_socket
AttributeError: 'NoneType' object has no attribute 'TLS_MODE'
```
Here is the relevant revised code:
```python
class _FakeSSLContext:
    def __init__(self, iface):
        self._iface = iface

    def wrap_socket(self, socket, server_hostname=None):
        """Return the same socket"""
        # pylint: disable=unused-argument


def set_socket(sock, iface=None):
    """Legacy API for setting the socket and network interface. Use a `Session` instead."""
    global _default_session  # pylint: disable=global-statement,invalid-name
    _default_session = Session(sock, _FakeSSLContext(iface))
    if iface:
        sock.set_interface(iface)
```
The `_FakeSSLContext` ends up with a `None` `iface`, and then `wrap_socket()` fails because `self._iface.TLS_MODE` will fail.

askpatrickw:
Fixes #48
Adds airlift config to example, and some HTTPS alongside the HTTP
Fixes all the pylint errors on the tests and breaks out the tests into their own pre-commit step to deal with duplicate code.

I am fixing this because the current Azure IOT examples using ESP32SPI connectivity fail without it.

@tannewt I have a question about my fix though as it also gives a pylint error

```
adafruit_requests.py:656:57:
W0212: Access to a protected member _the_interface of a client class (protected-access)
```

Am I misunderstanding your suggested fix in #48 ?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
     -   id: pylint
         name: pylint (library code)
         types: [python]
-        exclude: "^(docs/|examples/|setup.py$)"
+        exclude: "^(docs/|examples/|tests/|setup.py$)"
 -   repo: local
     hooks:
     -   id: pylint_examples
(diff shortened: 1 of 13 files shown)


tannewt:
Thanks @askpatrickw it looks good! Just disable pylint for the protected access.

askpatrickw:
Thanks @tannewt I just wanted to be sure.

ladyada:
fyi @brentru as well

tannewt (approved):
Thank you!

PR #87, opened 2021-12-09
Add type hints

tekktrik:
Probably the most complex typing PR I've done, so definitely would appreciate a thorough look from someone else @___@  I had to learn a few things for this one, so hopefully I implemented it all correctly!

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -37,6 +37,10 @@ __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Requests.git"
 
 import errno
+from ssl import SSLSocket
+from types import ModuleType
+
+from tests.mocket import SSLContext
 
 try:
     from typing import Union, TypeVar, Optional, Dict, Any, List
@@ -45,9 +49,14 @@ try:
     import adafruit_esp32spi.adafruit_esp32spi_socket as esp32_socket
     import adafruit_wiznet5k.adafruit_wiznet5k_socket as wiznet_socket
     import adafruit_fona.adafruit_fona_socket as cellular_socket
-    SocketType = TypeVar("SocketType", esp32_socket.socket, wiznet_socket.socket, cellular_socket.socket)
+    from adafruit_esp32spi.adafruit_esp32spi import ESP_SPIcontrol
+    from adafruit_wiznet5k.adafruit_wiznet5k import WIZNET5K
+    from adafruit_fona.adafruit_fona import FONA
+    import socket
+    SocketType = TypeVar("SocketType", esp32_socket.socket, wiznet_socket.socket, cellular_socket.socket, socket.socket)
     SocketpoolModuleType = TypeVar("SocketpoolModuleType", types.ModuleType("socket"), types.ModuleType("socketpool"))
     SSLContextType = TypeVar("SSLContextType", ssl.SSLContext) # Can use either CircuitPython or CPython ssl module
+    InterfaceType = TypeVar("InterfaceType", ESP_SPIcontrol, WIZNET5K, FONA)
 except ImportError:
     pass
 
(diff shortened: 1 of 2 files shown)


adafruit_requests.py
```diff
@@ -37,22 +37,51 @@
...
+    InterfaceType = TypeVar("InterfaceType", ESP_SPIcontrol, WIZNET5K, FONA)
+
+except ImportError:
+    pass
 
 # CircuitPython 6.0 does not have the bytearray.split method.
 # This function emulates buf.split(needle)[0], which is the functionality
 # required.
-def _buffer_split0(buf, needle):
+def _buffer_split0(buf, needle: bytes):
```
dhalbert, line 73: I think `bytes` is unnecessarily narrow here. You could use `_typing.ReadableBuffer`. Or maybe it is only `Union[bytes, bytearray]` ?

adafruit_requests.py
```diff
@@ -652,16 +710,18 @@ def connect(self, address):
...
+    def wrap_socket(
+        self, socket: SocketType, server_hostname: Optional[str] = None
+    ) -> _FakeSSLSocket:
         """Return the same socket"""
         # pylint: disable=unused-argument
         return _FakeSSLSocket(socket, self._iface.TLS_MODE)
 
 
-def set_socket(sock, iface=None):
+def set_socket(sock, iface: Optional[InterfaceType] = None) -> None:
```
dhalbert, line 724: What is the type of `sock`?

adafruit_requests.py
```diff
@@ -672,7 +732,15 @@ def set_socket(sock, iface=None):
     sock.set_interface(iface)
 
 
-def request(method, url, data=None, json=None, headers=None, stream=False, timeout=1):
+def request(
+    method,
```
dhalbert, line 736: What is the type of `method`?

adafruit_requests.py
```diff
@@ -187,7 +221,9 @@ def _readto(self, first, second=b""):
 
         return b""
 
-    def _read_from_buffer(self, buf=None, nbytes=None):
+    def _read_from_buffer(
+        self, buf: Optional[bytearray] = None, nbytes: Optional[int] = None
```
dhalbert, line 225: Could `buf` be something else besides a `bytearray` ?

dhalbert (changes requested):
Thanks for this difficult annotation. Just a few q's.

adafruit_requests.py
tekktrik: I've had issues with importing `_typing` if I remember correctly, but also `Union[bytes, bytearray]` seemed more like what it was intended for

adafruit_requests.py
tekktrik: `sock` is actually a module, either CircuitPython's `socketpool` or CPython's `socket` module.  Basically, it's any module which implements the `socket` class.

adafruit_requests.py
tekktrik: Also good catch!  I must missed this.  If I remember correctcly, `method` is a string defining the request method (GET, PUT, POST, etc)

adafruit_requests.py
tekktrik: I think this has to be of type `bytearray` based on this line:

https://github.com/tekktrik/Adafruit_CircuitPython_Requests/blob/72a300f9e339f4c9cd2e2364517bbf9aa997474d/adafruit_requests.py#L234

tekktrik:
Change # 1 - Changed to `Union[bytes, bytearray]` and also added the missing type hint for `buf`
Change # 2 - Added
Change # 3 - Added
Change # 4 - Left as `bytearray` for now but I'll change it if I misunderstood and that parameter can be of type `bytes`

dhalbert (approved):
Thanks for the iteration!

PR #91, opened 2021-12-28
From types in the try/catch clause

Neradoc:
Fixes:
```py
>>> import adafruit_requests
  File "code.py", line 1, in <module>
  File "adafruit_requests.py", line 40, in <module>
ImportError: no module named 'types'
```
(Issue reported on discord with 1.10.3)


dhalbert (approved):
Thanks for this fix!

PR #94, opened 2022-01-06
Fix importing `adafruit_requests` in CPython

### LINKED ISSUE #93: Importing versions >=1.10.3 broken in CPython
Some of the imports introduced in #87 (released in version 1.10.3) appear to break the import of `adafruit_requests` in CPython due to those new imports not being part of the `adafruit_requests` package's requirements.

To reproduce, you can create a fresh virtual environment and install `adafruit-circuitpython-requests==1.10.4`:
```
kevin@penguin:~$ python3 -m venv repro_venv
kevin@penguin:~$ . ./repro_venv/bin/activate
(repro_venv) kevin@penguin:~$ pip install adafruit-circuitpython-requests==1.10.4
Collecting adafruit-circuitpython-requests==1.10.4
  Downloading adafruit-circuitpython-requests-1.10.4.tar.gz (39 kB)
Collecting Adafruit-Blinka
  Downloading Adafruit-Blinka-6.17.0.tar.gz (153 kB)
     |████████████████████████████████| 153 kB 2.8 MB/s 
Collecting Adafruit-PlatformDetect>=3.13.0
  Downloading Adafruit-PlatformDetect-3.18.0.tar.gz (31 kB)
Collecting Adafruit-PureIO>=1.1.7
  Downloading Adafruit_PureIO-1.1.9.tar.gz (26 kB)
Collecting pyftdi>=0.40.0
  Downloading pyftdi-0.53.3-py3-none-any.whl (141 kB)
     |████████████████████████████████| 141 kB 13.6 MB/s 
Collecting pyserial>=3.0
  Downloading pyserial-3.5-py2.py3-none-any.whl (90 kB)
     |████████████████████████████████| 90 kB 3.1 MB/s 
Collecting pyusb!=1.2.0,>=1.0.0
  Downloading pyusb-1.2.1-py3-none-any.whl (58 kB)
     |████████████████████████████████| 58 kB 1.5 MB/s 
Using legacy 'setup.py install' for adafruit-circuitpython-requests, since package 'wheel' is not installed.
Using legacy 'setup.py install' for Adafruit-Blinka, since package 'wheel' is not installed.
Using legacy 'setup.py install' for Adafruit-PlatformDetect, since package 'wheel' is not installed.
Using legacy 'setup.py install' for Adafruit-PureIO, since package 'wheel' is not installed.
Installing collected packages: pyusb, pyserial, pyftdi, Adafruit-PureIO, Adafruit-PlatformDetect, Adafruit-Blinka, adafruit-circuitpython-requests
    Running setup.py install for Adafruit-PureIO ... done
    Running setup.py install for Adafruit-PlatformDetect ... done
    Running setup.py install for Adafruit-Blinka ... done
    Running setup.py install for adafruit-circuitpython-requests ... done
Successfully installed Adafruit-Blinka-6.17.0 Adafruit-PlatformDetect-3.18.0 Adafruit-PureIO-1.1.9 adafruit-circuitpython-requests-1.10.4 pyftdi-0.53.3 pyserial-3.5 pyusb-1.2.1
```

And then try importing `adafruit_requests` in the interpreter:

```python
(repro_venv) kevin@penguin:~$ python
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import adafruit_requests
  File "/home/kevin/repro_venv/lib/python3.9/site-packages/adafruit_requests.py", line 107, in <module>
    class Response:
  File "/home/kevin/repro_venv/lib/python3.9/site-packages/adafruit_requests.py", line 114, in Response
    def __init__(self, sock: SocketType, session: Optional["Session"] = None) -> None:
NameError: name 'SocketType' is not defined
```

I believe one of the imports of `adafruit_esp32spi`, `adafruit_wiznet5k`, etc. in this `try` block fail with `ImportError` since they are not requirements of `adafruit-circuitpython-requests` and thus don't get installed by `pip`, so the `SocketType` type var does not get defined (hence the error above):

https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/26b2411780f69edece156c25476359242446bac0/adafruit_requests.py#L41-L68

but since the `try` block's exception handler `pass`es on `ImportError`, we don't see an error until the first attempted usage of `SocketType` in this type annotation:

https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/26b2411780f69edece156c25476359242446bac0/adafruit_requests.py#L114

I read on Discord someone else ran into this as well: https://discord.com/channels/327254708534116352/537365702651150357/925282376525959229

Since those `adafruit_esp32spi`, `adafruit_wiznet5k`, etc. packages don't seem to be used beyond defining the `SocketType`, `InterfaceType`, etc. `TypeVar`s, could [protocols](https://docs.python.org/3/library/typing.html#typing.Protocol) be used to define the shape of those types to avoid needing to import those packages?

kevincon:
This PR fixes importing `adafruit_requests` in CPython by replacing the imports of various classes from `adafruit_esp32spi`, `adafruit_wiznet5k`, etc. (which weren't specified as requirements of `adafruit_requests` and thus caused importing `adafruit_requests` to fail due to not being available) with [protocols](https://mypy.readthedocs.io/en/stable/protocols.html) that describe the common structure of those types.

To help prevent the use of `adafruit_requests` in CPython from breaking in the future, this PR also adds a step to the Build CI job which uses tox to build the package and run its tests in a virtual environment.

I also tested the changes to `adafruit_requests.py` in CircuitPython by copying the file to the `lib/` folder of my MagTag device running the [COVID tracking project](https://learn.adafruit.com/magtag-covid-tracking-project-iot-display) and verifying that the display showed the COVID stats as expected (although note that project's data source [has stopped reporting new data since March 7, 2021](https://covidtracking.com/analysis-updates/giving-thanks-and-looking-ahead-our-data-collection-work-is-done)).

Fixes #93.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,10 +22,10 @@ jobs:
         awk -F '\/' '{ print tolower($2) }' |
         tr '_' '-'
         )
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
     - name: Versions
       run: |
         python3 --version
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,70 +10,70 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - name: Dump GitHub context
-        env:
-          GITHUB_CONTEXT: ${{ toJson(github) }}
-        run: echo "$GITHUB_CONTEXT"
-      - name: Translate Repo Name For Build Tools filename_prefix
-        id: repo-name
-        run: |
-          echo ::set-output name=repo-name::$(
-          echo ${{ github.repository }} |
-          awk -F '\/' '{ print tolower($2) }' |
-          tr '_' '-'
-          )
-      - name: Set up Python 3.7
-        uses: actions/setup-python@v1
-        with:
-          python-version: 3.7
-      - name: Versions
-        run: |
-          python3 --version
-      - name: Checkout Current Repo
-        uses: actions/checkout@v1
-        with:
-          submodules: true
-      - name: Checkout tools repo
-        uses: actions/checkout@v2
-        with:
-          repository: adafruit/actions-ci-circuitpython-libs
-          path: actions-ci
-      - name: Install dependencies
-        # (e.g. - apt-get: gettext, etc; pip: circuitpython-build-tools, requirements.txt; etc.)
-        run: |
-          source actions-ci/install.sh
-      - name: Pip install Sphinx, pre-commit
-        run: |
-          pip install --force-reinstall Sphinx sphinx-rtd-theme pre-commit
-      - name: Library version
-        run: git describe --dirty --always --tags
-      - name: Pre-commit hooks
-        run: |
-          pre-commit run --all-files
-      - name: Build assets
-        run: circuitpython-build-bundles --filename_prefix ${{ steps.repo-name.outputs.repo-name }} --library_location .
-      - name: Archive bundles
-        uses: actions/upload-artifact@v2
-        with:
-          name: bundles
-          path: ${{ github.workspace }}/bundles/
-      - name: Build docs
-        working-directory: docs
-        run: sphinx-build -E -W -b html . _build/html
-      - name: Check For setup.py
-        id: need-pypi
-        run: |
-          echo ::set-output name=setup-py::$( find . -wholename './setup.py' )
-      - name: Build Python package
-        if: contains(steps.need-pypi.outputs.setup-py, 'setup.py')
-        run: |
-          pip install --upgrade setuptools wheel twine readme_renderer testresources
-          python setup.py sdist
-          python setup.py bdist_wheel --universal
-          twine check dist/*
-      - name: Test Python package
-        run: |
-          pip install tox==3.24.5
-          tox
-      - name: Setup problem matchers
-        uses: adafruit/circuitpython-action-library-ci-problem-matchers@v1
+    - name: Dump GitHub context
+      env:
+        GITHUB_CONTEXT: ${{ toJson(github) }}
+      run: echo "$GITHUB_CONTEXT"
+    - name: Translate Repo Name For Build Tools filename_prefix
+      id: repo-name
+      run: |
+        echo ::set-output name=repo-name::$(
+        echo ${{ github.repository }} |
+        awk -F '\/' '{ print tolower($2) }' |
+        tr '_' '-'
+        )
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Versions
+      run: |
+        python3 --version
+    - name: Checkout Current Repo
+      uses: actions/checkout@v1
+      with:
+        submodules: true
+    - name: Checkout tools repo
+      uses: actions/checkout@v2
+      with:
+        repository: adafruit/actions-ci-circuitpython-libs
+        path: actions-ci
+    - name: Install dependencies
+      # (e.g. - apt-get: gettext, etc; pip: circuitpython-build-tools, requirements.txt; etc.)
+      run: |
+        source actions-ci/install.sh
+    - name: Pip install Sphinx, pre-commit
+      run: |
+        pip install --force-reinstall Sphinx sphinx-rtd-theme pre-commit
+    - name: Library version
+      run: git describe --dirty --always --tags
+    - name: Pre-commit hooks
+      run: |
+        pre-commit run --all-files
+    - name: Build assets
+      run: circuitpython-build-bundles --filename_prefix ${{ steps.repo-name.outputs.repo-name }} --library_location .
+    - name: Archive bundles
+      uses: actions/upload-artifact@v2
+      with:
+        name: bundles
+        path: ${{ github.workspace }}/bundles/
+    - name: Build docs
+      working-directory: docs
+      run: sphinx-build -E -W -b html . _build/html
+    - name: Check For setup.py
+      id: need-pypi
+      run: |
+        echo ::set-output name=setup-py::$( find . -wholename './setup.py' )
+    - name: Build Python package
+      if: contains(steps.need-pypi.outputs.setup-py, 'setup.py')
+      run: |
+        pip install --upgrade setuptools wheel twine readme_renderer testresources
+        python setup.py sdist
+        python setup.py bdist_wheel --universal
+        twine check dist/*
+    - name: Test Python package
+      run: |
+        pip install tox==3.24.5
+        tox
+    - name: Setup problem matchers
+      uses: adafruit/circuitpython-action-library-ci-problem-matchers@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,66 +10,70 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-    - name: Dump GitHub context
-      env:
-        GITHUB_CONTEXT: ${{ toJson(github) }}
-      run: echo "$GITHUB_CONTEXT"
-    - name: Translate Repo Name For Build Tools filename_prefix
-      id: repo-name
-      run: |
-        echo ::set-output name=repo-name::$(
-        echo ${{ github.repository }} |
-        awk -F '\/' '{ print tolower($2) }' |
-        tr '_' '-'
-        )
-    - name: Set up Python 3.7
-      uses: actions/setup-python@v1
-      with:
-        python-version: 3.7
-    - name: Versions
-      run: |
-        python3 --version
-    - name: Checkout Current Repo
-      uses: actions/checkout@v1
-      with:
-        submodules: true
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -10,66 +10,70 @@ jobs:
...
+        if: contains(steps.need-pypi.outputs.setup-py, 'setup.py')
+        run: |
+          pip install --upgrade setuptools wheel twine readme_renderer testresources
+          python setup.py sdist
+          python setup.py bdist_wheel --universal
+          twine check dist/*
+      - name: Test Python package
+        run: |
+          pip install tox==3.24.5
+          tox
```
kevincon, line 77: These lines are the only parts of this file I meant to modify (add), but unless you use GitHub's "hide whitespace" option when looking at the diff, it looks like every line in the file was modified. My guess is the file may have been committed with Windows line endings previously, so when I saved the file in Linux all of the line endings got replaced.

adafruit_requests.py
```diff
@@ -37,35 +37,117 @@
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Requests.git"
 
 import errno
+import sys
+
+if sys.implementation.name == "circuitpython":
```
kevincon, line 42: Previously this pattern was used to only run the typing-related code in Python implementations which support it (such as CPython):

```python
try:
    # typing-related code here
    ...
except ImportError:
    pass
```

But that caused the CPython import failure issue described in #93 to be swallowed, so to help make debugging similar issues easier in the future, I replaced that pattern with this approach of reading `sys.implementation.name` to check if CircuitPython is being used or not instead.

adafruit_requests.py
```diff
@@ -37,35 +37,117 @@
...
+    def cast(_t, value):
+        """No-op shim for the typing.cast() function which is not available in CircuitPython."""
+        return value
+
+
+else:
+    from ssl import SSLContext
+    from types import ModuleType, TracebackType
+    from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
+    from typing_extensions import Protocol
```
kevincon, line 53: `Protocol` was first introduced in Python 3.8's `typing` module, so since it seems `adafruit-requests` aims to support Python 3.7 (as seen by the Build CI job steps) and older versions (as seen by the classifiers used in `setup.py`), I added [`typing_extensions`](https://github.com/python/typing/tree/master/typing_extensions) as a dependency to `adafruit_requests` since it provides a backport of `Protocol` for Python versions <3.8.

adafruit_requests.py
```diff
@@ -37,35 +37,117 @@
...
+
+    SocketpoolModuleType = ModuleType
+
+    class InterfaceType(Protocol):
+        """Describes the structure every interface type must have."""
+
+        @property
+        def TLS_MODE(self) -> int:  # pylint: disable=invalid-name
+            """Constant representing that a socket's connection mode is TLS."""
+            ...
```
kevincon, line 147: I discovered CircuitPython does not appear to support class variables, because previously I wrote this `Protocol` class using this more concise form:

```python
class InterfaceType(Protocol):
    """Describes the structure every interface type must have."""
    
    TLS_MODE: int
```
but that resulted in a `SyntaxError: invalid syntax` exception being raised for the `TLS_MODE: int` line.

Luckily `Protocol` supports [an alternative syntax for these kinds of variables (specifically for Python 2.7-3.5)](https://www.python.org/dev/peps/pep-0544/#using-protocols-in-python-2-7-3-5) using `property` which I was able to leverage here.

docs/conf.py
```diff
@@ -25,7 +25,7 @@
 # Uncomment the below if you use native CircuitPython modules such as
 # digitalio, micropython and busio. List the modules you use. Without it, the
 # autodoc module docs will fail to generate with a warning.
-autodoc_mock_imports = ["adafruit_esp32spi", "adafruit_wiznet5k", "adafruit_fona"]
+# autodoc_mock_imports = ["digitalio", "busio"]
```
kevincon, line 28: Since we no longer use those imports, I put back what this was before #87.

tox.ini
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: 2022 Kevin Conley
+#
+# SPDX-License-Identifier: MIT
+
+[tox]
+envlist = py37
+
+[testenv]
+changedir = {toxinidir}/tests
```
kevincon, line 9: I included this `changedir` so that we run the tests from the `tests/` folder to help ensure that `adafruit_requests` gets imported from the installed package in tox's virtual environment instead of being imported locally from the root of the project folder. I believe that should result in the tests importing `adafruit_requests` in a manner more representative of how CPython users would if they installed the package via `pip`.

adafruit_requests.py
```diff
@@ -157,7 +239,7 @@ def _recv_into(self, buf: bytearray, size: int = 0) -> int:
             read_size = len(b)
             buf[:read_size] = b
             return read_size
-        return self.socket.recv_into(buf, size)
+        return cast("SupportsRecvInto", self.socket).recv_into(buf, size)
```
kevincon, line 242: This `cast()` is needed for mypy to understand that we can safely call `recv_into()` on `self.socket` here because the legacy type of CircuitPython socket (represented by the `LegacyCircuitPythonSocketType` and the `self._backwards_compatible` flag checked above being `True`) does not have a `recv_into()` method.

adafruit_requests.py
```diff
@@ -440,7 +522,7 @@ def _free_sockets(self) -> None:
 
     def _get_socket(
         self, host: str, port: int, proto: str, *, timeout: float = 1
-    ) -> SocketType:
+    ) -> CircuitPythonSocketType:
```
kevincon, line 525: This change was needed because the `recv_into()` method is called on the socket returned by this `_get_socket()` method in the `request()` method, and not every `SocketType` provides a `recv_into()` method (specifically `LegacyCircuitPythonSocketType` does not). `CircuitPythonSocketType` meets that requirement while also being a valid `SocketType` to support the socket being passed to other methods which require a `SocketType` socket.

.github/workflows/build.yml
Neradoc: It's the indentation that have been changed.
Every line after `steps:` has 2 additional spaces.

kevincon: Ah good catch, I think VS Code must have done that automatically. I'll revert that.

adafruit_requests.py
dhalbert: We can move the requirements up to 3.8. The 3.7 choice was made years ago.

kevincon: @dhalbert sure I can do that in this PR.

Are these the changes you would expect? Am I missing anything?

- [ ] Add `python_requires='>=3.8'` to `setup.py`
- [ ] Replace 3.4/3.5 classifiers in `setup.py` with a 3.8 classifier
- [ ] Update `.readthedocs.yaml` to use 3.8 instead of 3.7
- [ ] Update `tox.ini` to use py38 instead of py37
- [ ] Update `build.yml` to use 3.8 instead of 3.7 to run `test` job steps
- [ ] Update `release.yml` to use 3.8 instead of 3.x to run `upload-pypi` job steps
- [ ] Import `Protocol` from `typing` instead of `typing_extensions`
- [ ] Remove `typing-extensions` from `requirements.txt` and `setup.py`'s `install_requires` list

FoamyGuy: @kevincon This list looks good to me. Go ahead and add commit(s) with these changes when you have the opportunity. Thank you for working on this!

FoamyGuy:
I successfully tested the version from this branch on PyPortal and CPython with the simpletest scripts in the repo. 

The changes you've made and reasoning for them make sense to me. Thanks again for working on this.

FoamyGuy (approved):
Re-tested with latest commits on PyPortal and CPython both still working as expected. 

Changes look good to me. Thank you @kevincon

tekktrik:
This is great!  The only question I have is how often this library gets used on CPython w/ Blinka if the CPython `requests` module exists?  Just trying to understand use cases.

tekktrik:
Never mind, I caught up on the relevant Issue.  Sounds like a cool project!

adafruit_requests.py
FoamyGuy: @kevincon it seems that the change here caused the library to stop working on microcontrollers with this error:
```
code.py output:
Connecting to AP...
Connected to WBurn-2G 	RSSI: -50
Fetching text from http://wifitest.adafruit.com/testwifi/index.html
  File "code.py", line 58, in <module>
  File "adafruit_requests.py", line 847, in get
  File "adafruit_requests.py", line 719, in request
  File "adafruit_requests.py", line 210, in __init__
  File "adafruit_requests.py", line 297, in _readto
  File "adafruit_requests.py", line 241, in _recv_into
TypeError: function takes 2 positional arguments but 3 were given

Code done running.
```
I think we'll need to revert it for now. Curious if you have any ideas that could make it worth in both environments?

FoamyGuy: Actually maybe this is not what is causing the error. My mistake. I reverted it in my local copy, but still seeing that error

PR #95, opened 2022-01-09
fallback types

FoamyGuy:
If certain imports used for typing are missing, these custom TypeVars can end up not being defined which causes sphinx to fail to build with them ex: https://github.com/adafruit/Adafruit_CircuitPython_FunHouse/pull/28/checks

This change provides minimal fallbacks for those types so that sphinx build will complete in this case.

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -65,7 +65,11 @@ try:
     InterfaceType = TypeVar("InterfaceType", ESP_SPIcontrol, WIZNET5K, FONA)
 
 except ImportError:
-    pass
+    # pylint: disable=invalid-name
+    SocketType = None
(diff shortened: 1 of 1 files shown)


anecdata (commented):
Regression-tested in a couple of typical GET/POST applications. No change to operation, as expected.

FoamyGuy:
Closing this. As discussed "in the weeds" during the meeting 1/10/21 the plan will be to remove the requirements and define interfaces for the types that are coming from them.

tekktrik:
Thanks for posting this!  I think I missed that update, so I'll go back and listen to it, thanks for that!  is there any issue for going back and updating it when it's time?

PR #102, opened 2022-03-21
Remove pre 7.0 compatibility; other cleanup

dhalbert:
Simplify and shorten the current code. I will have a further PR later that changes a bunch of thrown exceptions to allow for easier error handling.

- CircuitPython 7 supports `split()` on bytes and bytearrays. Remove hand-coded routines no longer needed.
- CircuitPython 6 supports streaming `json` decoding. Remove workaround for 5.x non-streaming decoding.
- Refactor the duplicated check for gzip content.

_Please_ test with your own wifi code - thanks. I tested this with the simple wifi testing code, and with the Google calendar example, but not with other examples.

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -45,7 +45,6 @@ if sys.implementation.name == "circuitpython":
         """No-op shim for the typing.cast() function which is not available in CircuitPython."""
         return value
 
-
 else:
     from ssl import SSLContext
     from types import ModuleType, TracebackType
@@ -148,16 +147,6 @@ else:
     SSLContextType = Union[SSLContext, "_FakeSSLContext"]
 
 
-# CircuitPython 6.0 does not have the bytearray.split method.
-# This function emulates buf.split(needle)[0], which is the functionality
-# required.
-def _buffer_split0(buf: Union[bytes, bytearray], needle: Union[bytes, bytearray]):
-    index = buf.find(needle)
-    if index == -1:
-        return buf
-    return buf[:index]
-
-
 class _RawResponse:
     def __init__(self, response: "Response") -> None:
         self._response = response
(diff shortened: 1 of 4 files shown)


anecdata (changes requested):
Running fine so far on PyPortal.

Getting some of these on Metro ESP32-S2:
```
  File "adafruit_requests.py", line 727, in post
  File "adafruit_requests.py", line 667, in request
  File "adafruit_requests.py", line 578, in _send_request
  File "adafruit_requests.py", line 554, in _send
NameError: name 'EAGAIN' is not defined
```

dhalbert:
Re-pushed; due to git problems, the code that was pushed was wrong.

anecdata:
Passes initial testing with my "production" code on PyPortal and Metro ESP32-S2 (CircuitPython 7.2.0). I'll load it on more devices and let it run.

askpatrickw:
I ran pytest locally and hit the same failure as the CI.

Changing https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/main/tests/post_test.py#L83
to `sock.send.assert_called_with(b'Date=July 25, 2019')`

Fixes the test... well it passed. But I don't understand why. ;-)

adafruit_requests.py
```diff
@@ -637,11 +589,9 @@ def _send_request(
...
             assert data is None
             # pylint: disable=import-outside-toplevel
-            try:
-                import json as json_module
-            except ImportError:
-                import ujson as json_module
-            data = json_module.dumps(json)
+            import json
+
+            data = json.dumps(json)
```
Neradoc, line 594: This will fix the error detected by pytest.
```suggestion
            import json as json_module

            data = json_module.dumps(json)
```
Too many jsons in the kitchen !

Neradoc (changes requested):
That's why ;)

adafruit_requests.py
dhalbert: Thanks! I will finish this today (will be out this morning). Yes, we could just import it at the top. I'm not sure why there was a conditional import. Since it's built-in, the size is minimal.

dhalbert:
Also tested with a PyPortal (ESP32 coprocessor) with the OpenWeather demo. Ready for re-review.

anecdata (commented):
Earlier version (w/o the `json` fix) has been running for a day on two types of 'production' code on Airlift and ESP32-S2/3 without issue. Expanded the test with current library to more applications, no issues found.

PR #107, opened 2022-03-29
Use typing_extensions.Protocol if needed

tekktrik:
New black hated that space apparently 😖 

Also took the opportunity to resolve #107

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -52,10 +52,10 @@ else:
     from types import ModuleType, TracebackType
     from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
 
-try:
-    from typing import Protocol
-except ImportError:
-    from typing_extensions import Protocol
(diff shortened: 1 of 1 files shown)


adafruit_requests.py
```diff
@@ -47,11 +47,15 @@ def cast(_t, value):
...
 else:
     from ssl import SSLContext
     from types import ModuleType, TracebackType
-    from typing import Any, Dict, List, Optional, Protocol, Tuple, Type, Union, cast
+    from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
+
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol
```
dhalbert, line 58: Make this part of the `else` (indent it one level), since it's type-checking only.

dhalbert (changes requested):

Issue #103, opened 2022-03-21
Getting a request.py error in the code for the "MagTag Lists From Google Spreadsheets" tutorial


wavesailor:
I'm getting an error running the following code: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/MagTag_Google_Sheets/weekly_planner/code.py
```
Connecting to AP MyWiFI
Updating time
Getting time for timezone America/New_York
Updating tasks
Retrieving data...Traceback (most recent call last):
  File "code.py", line 101, in <module>
  File "adafruit_portalbase/network.py", line 478, in fetch
  File "adafruit_requests.py", line 769, in get
  File "adafruit_requests.py", line 758, in request
  File "adafruit_requests.py", line 702, in request
  File "adafruit_requests.py", line 378, in close
ValueError: invalid syntax for integer with base 16

Code done running.
```

I use `wget` on my Pi  to retrieve my URL just fine, but it does not work on the MagTag

I'm using the latest libraries and Circuit Python for MagTag
```
adafruit-circuitpython-adafruit_magtag_2.9_grayscale-en_US-7.2.3.uf2
adafruit-circuitpython-bundle-7.x-mpy-20220321.zip
```

wavesailor:
I kinda follow what you saying .... my initial request is redirected (Temporary redirect 307) to another host but still Google.

But does this mean `requests.py` needs to be corrected/fixed?

anecdata:
I believe `adafruit_requests` would need a change, since the host changes on redirect. You _might_ be able to put the final `https://doc-14-2g-sheets.googleusercontent.com/`... URL into the CircuitPython code for it to work. It's a temporary redirect, so it could change again in the future.

wavesailor:
Okay thanks. Do I have to log a request somewhere to get it fixed?

PS. I did try use the redirected URL in but it expires are a certain period  :-(

anecdata:
Yeah, I couldn't get it to work in curl either. A bit unfriendly to require a heavy redirect to access an asset.

This issue serves as the request to get the issue addressed, but it is dependent on resources and priorities, or someone motivated enough to fix it sooner ;-). I'm going to also file a new issue in the Guide to keep track there... maybe there's an alternate way to serve up the spreadsheet.

anecdata:
@wavesailor It might work to get the response headers and parse out the new temporary `Location` URL and then turn around and fetch that.

wavesailor:
[...]

PR #110, opened 2022-07-03
Proposed fix for issue #103. Fixes an issue with _throw_away() and large chunked redirects.

klocs:
I found that `Reponse._throw_away()` worked fine unless the bytes returned from `_recv_into()` wasn't the full size of the throw away buffer. Simple fix once I figured out what was happening. Also including a unit test the reproduces the problem.

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -321,12 +321,12 @@ class Response:
         for _ in range(nbytes // len_buf):
             read = 0
             while read < len_buf:
-                read += self._recv_into(buf, len_buf-read)
+                read += self._recv_into(buf, len_buf - read)
         remaining = nbytes % len_buf
         if remaining:
             read = 0
             while read < remaining:
-                read += self._recv_into(buf, remaining-read)
+                read += self._recv_into(buf, remaining - read)
 
     def close(self) -> None:
         """Drain the remaining ESP socket buffers. We assume we already got what we wanted."""
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -319,9 +319,9 @@ class Response:
         buf = self._receive_buffer
         len_buf = len(buf)
         for _ in range(nbytes // len_buf):
-            read = 0
-            while read < len_buf:
-                read += self._recv_into(buf, len_buf - read)
+            to_read = len_buf
+            while to_read > 0:
+                to_read -= self._recv_into(buf, to_read)
         remaining = nbytes % len_buf
         if remaining:
             read = 0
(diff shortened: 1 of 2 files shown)


adafruit_requests.py
```diff
@@ -317,11 +317,16 @@ def _throw_away(self, nbytes: int) -> None:
...
 
         buf = self._receive_buffer
-        for _ in range(nbytes // len(buf)):
-            self._recv_into(buf)
-        remaining = nbytes % len(buf)
+        len_buf = len(buf)
+        for _ in range(nbytes // len_buf):
+            read = 0
+            while read < len_buf:
+                read += self._recv_into(buf, len_buf - read)
```
dhalbert, line 324: How about this, which I think is a little clearer, and does less arithmetic:
```suggestion
            to_read = len_buf
            while to_read > 0:
                to_read -= self._recv_into(buf, to_read)
```

adafruit_requests.py
```diff
@@ -317,11 +317,16 @@ def _throw_away(self, nbytes: int) -> None:
...
+        remaining = nbytes % len_buf
         if remaining:
-            self._recv_into(buf, remaining)
+            read = 0
+            while read < remaining:
+                read += self._recv_into(buf, remaining - read)
```
dhalbert, line 329:
```suggestion
            while remaining > 0:
                remaining -= self._recv_into(buf, remaining)
```

dhalbert (changes requested):
Thanks for figuring this out! I have two related coding suggestions.

adafruit_requests.py
klocs: I like it.

adafruit_requests.py
```diff
@@ -324,9 +324,8 @@ def _throw_away(self, nbytes: int) -> None:
                 to_read -= self._recv_into(buf, to_read)
         remaining = nbytes % len_buf
         if remaining:
```
klocs, line 326: Looking at it again, I think we can eliminate this `if remaining:` check also. What do you think?
I'm pretty new at working with github's UI so I'm not sure if I'm doing this right.

dhalbert (approved):

adafruit_requests.py
dhalbert: You're right, it's redundant! I made a release but go ahead and make another PR, and I'll make another reelase.

dhalbert: maybe `to_read` and `remaining` could have the same name.

klocs: Sounds good. I'll work on it and see if it makes sense to combine the variables. It might be confusing though. I'll send another pull request in a bit, may not be today.

PR #118, opened 2022-10-27
Simple async http requests

alexmherrmann:
This will never go anywhere unless there's a pull request but I know this one isn't great yet, happy to work on it when possible as I think it's useful.

I took some of @dhalbert 's code that made async requests a thing and tried to write a small test and pull out some other experimental changes made. It won't do much until the socket read is also async but that is beyond my pay grade at the moment.

### DIFF
--- a/adafruit_async_requests.py
+++ b/adafruit_async_requests.py
@@ -1,0 +1,234 @@
+# SPDX-FileCopyrightText: 2020 Dan Halbert for Adafruit Industries
+# SPDX-FileContributor: Updated and repackaged/tested by Alex Herrmann, 2022
+#
+# SPDX-License-Identifier: MIT
+
+"""
+`adafruit_requests.async_session`
+================================================================================
+
+"""
+
+import errno
+
+import json as json_module
+
+import asyncio
+from adafruit_requests import Session, Response, OutOfRetries
+
+try:
+    from typing import Any, Dict, List, Optional
+    from circuitpython_typing.socket import (
+        SocketType,
+        SocketpoolModuleType,
+        SSLContextType,
+    )
+except ImportError:
+    pass
+
+
+class AsyncSession(Session):
+    """HTTP session that shares sockets and ssl context."""
+
+    def __init__(
+            self,
+            socket_pool: SocketpoolModuleType,
+            ssl_context: Optional[SSLContextType] = None,
+    ) -> None:
+        Session.__init__(socket_pool, ssl_context)
+        self._last_response = None
+
+    @staticmethod
+    async def _asend(socket: SocketType, data: bytes):
+        total_sent = 0
+        while total_sent < len(data):
+            # ESP32SPI sockets raise a RuntimeError when unable to send.
+            try:
+                sent = socket.send(data[total_sent:])
+            except OSError as exc:
+                if exc.errno == errno.EAGAIN:
+                    # Can't send right now (e.g., no buffer space), try again.
+                    await asyncio.sleep(0)
+                # Some worse error.
+                raise
+            except RuntimeError as exc:
+                raise OSError(errno.EIO) from exc
+            if sent is None:
+                sent = len(data)
+            if sent == 0:
+                # Not EAGAIN; that was already handled.
+                raise OSError(errno.EIO)
+            total_sent += sent
+
+
+    async def _asend_request(
+            self,
+            socket: SocketType,
+            host: str,
+            method: str,
+            path: str,
+            headers: List[Dict[str, str]],
+            data: Any,
+            json: Any,
+    ):
+        # pylint: disable=too-many-arguments
+        await self._asend(socket, bytes(method, "utf-8"))
+        await self._asend(socket, b" /")
+        await self._asend(socket, bytes(path, "utf-8"))
+        await self._asend(socket, b" HTTP/1.1\r\n")
+        if "Host" not in headers:
+            await self._asend(socket, b"Host: ")
+            await self._asend(socket, bytes(host, "utf-8"))
+            await self._asend(socket, b"\r\n")
+        if "User-Agent" not in headers:
+            await self._asend(socket, b"User-Agent: Adafruit CircuitPython\r\n")
+        # Iterate over keys to avoid tuple alloc
+        for k in headers:
+            await self._asend(socket, k.encode())
+            await self._asend(socket, b": ")
+            await self._asend(socket, headers[k].encode())
+            await self._asend(socket, b"\r\n")
+        if json is not None:
+            assert data is None
+            data = json_module.dumps(json)
+            await self._asend(socket, b"Content-Type: application/json\r\n")
+        if data:
+            if isinstance(data, dict):
+                await self._asend(
+                    socket, b"Content-Type: application/x-www-form-urlencoded\r\n"
+                )
+                _post_data = ""
+                for k in data:
+                    _post_data = "{}&{}={}".format(_post_data, k, data[k])
+                data = _post_data[1:]
+            await self._asend(socket, b"Content-Length: %d\r\n" % len(data))
+        await self._asend(socket, b"\r\n")
+        if data:
+            if isinstance(data, bytearray):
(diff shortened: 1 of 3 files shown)


alexmherrmann:
I will merge in the original so hopefully the merge conflict goes away.

FoamyGuy:
In order for this to live in this repo it would need to be converted to a directory instead of single .py files. Some of the tooling that is used to generate library bundles and other things assumes that there will be only a single .py file, or a single directory containing the library .py files in the root of the repo. 

Another option could be to create a seperate repo for the async version, then it could live in it's own repo as a single file library.

alexmherrmann:
Thank you @FoamyGuy, I am very new to circuitpython obviously, would you recommend one way over the other? Perhaps @dhalbert has an opinion as it was originally his code.

Is there an example library with multiple .py files so I can get it right if it's the first way you mentioned?

tekktrik:
@FoamyGuy correct me if I'm wrong, but the decision was to have this been it's own library.  Not sure how we should start this process then with regards to this pull request

FoamyGuy:
That is correct. Sorry I meant to comment back here after it was discussed but didn't do it. 

The discussion is here: https://youtu.be/n8Y6tT6nSig?t=3032 and yep conses was ultimately to make a new library but be clear that it should keep the same functionality as this one, i.e. new things should get added to both instead of just one. 

@alexmherrmann If you want to take what you have and put it into a new repo to get us kicked off that would be awesome. If your up for it there is a tool called cookie-cutter that we use to create the new libraries, it and how to use it are documented here: https://learn.adafruit.com/creating-and-sharing-a-circuitpython-library/creating-a-library. If you'd rather not go through that process, that is okay as well. Make a new repo with the async code and ideally at least one example in it and link that here. We can run the cookie cutter and move your code into the generated library files afterward.

alexmherrmann:
Thank you all, I should have some time this upcoming week. I haven't quite
got the best way to write tests yet. Between the mocks and what I've seen
in other repository it is still a little foreign to me but I'll try my best.

There were a few places it could block in the code too that would be useful
to note. The socket read was the big one but I think @dhalbert had a way
around that.

Are there requirements around documentation? Or is that all in the cookie
cutter library too?

FoamyGuy:
@alexmherrmann cookie-cutter will prompt you with yes/no whether you want it to generate the docs files. All official libraries choose yes for that option. 

The content for the docs pages is taken from the docstrings in the code. We use a tool called sphinx to build it into html pages from those docstrings. This page talks about that process a bit: https://learn.adafruit.com/creating-and-sharing-a-circuitpython-library/sharing-our-docs-on-readthedocs and shows the command to run to make a build locally. Much of the page deals with setting up Readthe docs specifically which is something that an Adafruit team member would do once this repo is created, it requires permissions within a few different systems.

veloyage:
Is this still being worked on, by anyone? Is there no interest any more, or did circumstances change? I would be quite interested in async requests, as currently they hang my UI. Only for a second or so, but still ugly. As wifi becomes a default peripheral, and internet connected projects the norm, I imagine this will become a common annoyance. 

Wondering whether the library structure and cookiecutter stuff was the main issue, I tried to just run the existing code and ran into errors. Also, early in the thread @alexmherrmann said "It won't do much until the socket read is also async but that is beyond my pay grade at the moment." - that's not just above my paygrade, I am not even sure what that implies. Is significant heavy lifting required to make this useful?

Would be glad if someone with a better overview could enlighten me. I am happy to help, but clueless as to how.

dhalbert:
> "It won't do much until the socket read is also async but that is beyond my pay grade at the moment." - that's not just above my paygrade, I am not even sure what that implies. Is significant heavy lifting required to make this useful?

That is a significant addition - I think we need core CircuitPython work for that. A lot of the infrastructure is in place but it hasn't be done yet.

alexmherrmann:
Hello all, my job has changed up quite a bit for the better and I'm back around and able to start helping. I was going to get main merged and updates in to the branch to make sure what was working before is still working now. We're moving into a house so things will be chaotic but the magtag is getting plugged back in as soon as we settle!

What @dhalbert is referring to, and can speak much more confidently about is that we can add as much async code in as we want at this level. However, (as I understand it and please correct me 😭) because the socket receive is still synchronous it's not "really" async. It's been a while since I got into the code however.

@dhalbert That would mean modifying the core C code correct?

Sorry for the radio silence all ✌️

dhalbert:
> What @dhalbert is referring to, and can speak much more confidently about is that we can add as much async code in as we want at this level. However, (as I understand it and please correct me 😭) because the socket receive is still synchronous it's not "really" async. It's been a while since I got into the code however.
> 
> @dhalbert That would mean modifying the core C code corr

Right, exactly.

dhalbert:
A lot of my initial changes were a cleanup of the exceptions, etc., and that has been done, though in a different form. The core part are really `arequest()`, etc.

czei:
I am confused by "It won't do much until the socket read is also async".  Maybe this is older than the current API, but to someone new to CP it _looks_ like the socket class supports asynchronous reads:

https://docs.circuitpython.org/en/latest/shared-bindings/socketpool/index.html

My current project is being prevented from shipping because the entire device locks up while a large JSON file is being read, so I'm motivated to find a short-term solution.

strefli3:
I too have some motivation for async http requests. Is this pull request the best place for someone to start looking at solutions?

strefli3:
Here is the fix to my issue above

```
diff --git a/adafruit_async_requests.py b/adafruit_async_requests.py
index 35574d5..8b934cb 100644
--- a/adafruit_async_requests.py
+++ b/adafruit_async_requests.py
@@ -35,7 +35,7 @@ class AsyncSession(Session):
             socket_pool: SocketpoolModuleType,
             ssl_context: Optional[SSLContextType] = None,
     ) -> None:
-        Session.__init__(socket_pool, ssl_context)
+        Session(socket_pool, ssl_context)
 
         #FIXME Alex Herrmann: Don't know why the below are required to make TOX work
         self._socket_pool = socket_pool
```

I'll continue to hack on this [fork](https://github.com/strefli3/CircuitPython_AsyncRequests/tree/main-async).

The branch `main-async` is noticably faster - my LED array has went for a very noticeable pause to just a short blip.

strefli3:
In the spirit of adding more `await` and `async` into the code I have created 

* `AsyncSession`
* `AsyncResponse`
* `_AsyncRawResponse`

I am not sure if this provides a step in the right direction, but at least it facilitates a bit of momentum for the discussion.

To date the function `_async_readto` seems to take the most time which I have pasted here for discussion purposes. 

```
    async def _async_readto(self, stop: bytes) -> bytearray:



        """Async version of  _readto method


        


        This seems to be the function that takes the most "time".


        


        """


        buf = self._receive_buffer


        end = self._received_length


        while True:


            i = buf.find(stop, 0, end)


            if i >= 0:


                # Stop was found. Return everything up to but not including stop.


                result = buf[:i]


                new_start = i + len(stop)


                # Remove everything up to and including stop from the buffer.


                new_end = end - new_start


                buf[:new_end] = buf[new_start:end]


                self._received_length = new_end


                return result





            # Not found so load more bytes.


            # If our buffer is full, then make it bigger to load more.


            if end == len(buf):


                new_buf = bytearray(len(buf) + 32)


                new_buf[: len(buf)] = buf


                buf = new_buf


                self._receive_buffer = buf





            read = self._recv_into(memoryview(buf)[end:])


            if read == 0:


                self._received_length = 0


                return buf[:end]


            end += read
```

anecdata:
Just an aside... I did some timings long ago and one of the key culprits was the `if result == b"H":` in Requests. It tied the first part of the response to the request, but there was often a relatively large time between sending the last byte of the request and receiving the first "H" of the response (transit time + server time + transit time).

Another aside... do we have a feel for a good async granularity (async byte at a time, async chunk of size N bytes at a time, async request then async response)? One of the trade-offs I would think would be the performance hit of too-small chunks, vs. the async responsiveness to non-Requests parts of the application.

PR #124, opened 2023-01-09
preserve last exception when things are going wrong

jepler:
CircuitPython 8 supports exception chaining, so that the original problem can be shown. Here's how it looks on desktop python3:
```
  File "/home/jepler/src/bundle/libraries/helpers/requests/adafruit_requests.py", line 527, in _get_socket
    sock.connect((connect_host, port))
ConnectionRefusedError: [Errno 111] Connection refused

The above exception was the direct cause of the following exception:

  File "/home/jepler/src/bundle/libraries/helpers/requests/adafruit_requests.py", line 721, in get
    return self.request("GET", url, **kw)
  File "/home/jepler/src/bundle/libraries/helpers/requests/adafruit_requests.py", line 661, in request
    socket = self._get_socket(host, port, proto, timeout=timeout)
  File "/home/jepler/src/bundle/libraries/helpers/requests/adafruit_requests.py", line 508, in _get_socket
    raise RuntimeError("Sending request failed") from last_exc
RuntimeError: Sending request failed
```

Before, just the second block would be shown, and it would be hidden that the underlying cause was "connection refused".

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -499,19 +499,22 @@ class Session:
         )[0]
         retry_count = 0
         sock = None
+        last_exc = None
         while retry_count < 5 and sock is None:
             if retry_count > 0:
                 if any(self._socket_free.items()):
                     self._free_sockets()
                 else:
-                    raise RuntimeError("Sending request failed")
+                    raise RuntimeError("Sending request failed") from last_exc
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Thanks for this diagnostic help!

PR #129, opened 2023-03-09
Steam API example

DJDevon3:
API example displays the sum total amount of time in hours and days playing video games on steam.

### DIFF
--- a/examples/requests_api_discord.py
+++ b/examples/requests_api_discord.py
@@ -89,7 +89,6 @@ while True:
         # Print keys to Serial
         discord_debug_keys = True  # Set to True to print Serial data
         if discord_debug_keys:
-
             ada_discord_all_members = ada_res["approximate_member_count"]
             print("Members: ", ada_discord_all_members)
 
--- a/examples/requests_api_github.py
+++ b/examples/requests_api_github.py
@@ -81,7 +81,6 @@ while True:
         # Print Keys to Serial
         gh_debug_keys = True  # Set True to print Serial data
         if gh_debug_keys:
-
             github_id = github_response["id"]
             print("UserID: ", github_id)
 
--- a/examples/requests_api_mastodon.py
+++ b/examples/requests_api_mastodon.py
@@ -33,6 +33,7 @@ except ImportError:
     print("Secrets File Import Error")
     raise
 
+
 # Converts seconds in minutes/hours/days
 def time_calc(input_time):
     if input_time < 60:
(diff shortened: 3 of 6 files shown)


tekktrik:
I can take a look at this in a couple days, if no one does prior.

FoamyGuy (changes requested):
Code and functionality look good to me. 

I think it would be good to add a comment or two with info about how to get a steam API key and how to find the steam ID for your user account. Google got me to both items, but a point in the right direction from a comment in the example would be helpful.

I wish that their API returned more plain text rather than ID links to other objects. It'd be interesting to see the breakdown of game titles with hours played for each, but it seems like it'd require additional calls into their API to match up IDs to game titles. 

I don't think it matters too much for this example, but during testing I found that the API omits free games from the results. [This SO post](https://stackoverflow.com/questions/24463184/getownedgames-from-steamapi-not-working) shows a URL param that can be added to include them. Which in my case means the scary number that came out the first time wasn't even the real number, it got much scarier once it included the stats for one of my highest played games =x

DJDevon3:
That makes a lot of sense why my total game # was different from the amount of game titles it returned.  Added `&include_played_free_games=1` to the URL and instantly returned every game. That completely solved the discrepancy which I did notice and baffled me. Thanks!

As for showing time played for every title that's absolutely possible and rather easy. I bypassed that for the harder challenge which was taking every single game it spit out and summed the total time played from every game.

Like I said in the meeting, if you play a lot of games you might not want to see that number. 😆 

Added URL's to the API docs and key at least where it resides for now. One thing I've learned about API's is they'll change URL's sometimes so do you include it or not.  At least they'll work for the time being. Submitting changes in a commit/PR.

DJDevon3:
pylint & black hate me.  looks like that one went through.

DJDevon3:
Without free games included
```py
Attempting to GET STEAM Stats!
===============================
Total Games:  153
Total Hours: 5286.65
Total Days: 220.277
Monotonic:  2299.87

Finished!
Next Update in 15 minutes
```
With free games included
```py
Attempting to GET STEAM Stats!
===============================
Total Games:  156
Total Hours: 5320.8
Total Days: 221.7
Monotonic:  2379.57

Finished!
Next Update in 15 minutes
===============================
```

DJDevon3:
What would be truly scary is pulling the games, value, and spitting out a dollar amount. 😨

DJDevon3:
@tekktrik or @FoamyGuy Should be ready for another review or do I need to do something else?

FoamyGuy (approved):
Looks good to me. I re-tested the latest version on a Feather ESP32-S2 TFT.

Thank you @DJDevon3!

PR #131, opened 2023-04-21
Add docstrings for response.status_code and reason

Neradoc:
For RTD.


jposada202020 (approved):
LGTM. Thanks @Neradoc :)

PR #132, opened 2023-05-19
implement allow_redirects argument

FoamyGuy:
this change adds the `allow_redirects` argument to the request methods. It defaults to True so that current behavior stays the same for anyone who doesn't explicitly change the argument to false. 

If it's False it will return the redirect response (300 level status code) to the caller that initiated the request instead of automatically following the redirect and returning the final response to the caller.

This matches the functionality in the [CPython requests library API](https://requests.readthedocs.io/en/latest/user/quickstart/#redirection-and-history)

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -631,6 +631,7 @@ class Session:
         headers: Optional[Dict[str, str]] = None,
         stream: bool = False,
         timeout: float = 60,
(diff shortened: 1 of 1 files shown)


tekktrik (approved):
I didn't test but this looks sensible and correct to me, thanks for the addition!

PR #133, opened 2023-05-29
Add OpenSky-Network API example

DJDevon3:
Free public limit is 100 calls per day and maximum of 60 seconds each or you'll get temp banned. You get 4000 calls if you login but I haven't figured out how with a microcontroller yet. They have no token system and their API is primarily built for python & curl. It's a foot in the door for public access to flight data.

### DIFF
--- a/examples/requests_api_openskynetwork_private.py
+++ b/examples/requests_api_openskynetwork_private.py
@@ -1,0 +1,153 @@
+# SPDX-FileCopyrightText: 2023 DJDevon3
+# SPDX-License-Identifier: MIT
+# Coded for Circuit Python 8.1
+# DJDevon3 ESP32-S3 OpenSkyNetwork_Private_API_Example
+
+import os
+import time
+import ssl
+import json
+import wifi
+import socketpool
+import adafruit_requests
+import circuitpython_base64 as base64
+
+# OpenSky-Network.org Login required for this API
+# REST API: https://openskynetwork.github.io/opensky-api/rest.html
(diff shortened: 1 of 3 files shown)


DJDevon3:
Figured out how to authenticate a login. It was not easy and there was almost no documentation. Authentication via REST GET is undocumented. Just figured it out from hints reading their forums.

DJDevon3:
The private_area example is while being authenticated (use website login credentials) and shows all air traffic in a lat/lon geographic area. You're going to want to review this on an S2 or S3.  It will return a good amount of data depending on how large of a geographic area you select.  The default area I put in is the Orlando airport area so there's always bound to be some traffic happening.

OpenSky_Area_API

FoamyGuy (changes requested):
I tested all 3 examples successfully on a feather esp32-s3. 

I did a test run without those timezone variables and it was able to run and fetch data successfully. I think we can remove those in order to avoid any confusion of someone thinking they will have some impact on the results. 

Everything beyond that seems good to me. Thanks for the new examples @DJDevon3!

DJDevon3:
Oops yes. About halfway through became unsure if everything should be in 0 UTC or add the option for timezone offsets.  I think all flight times are returned as 0 UTC because flights can easily cross multiple timezones.  Easier keep everything in 0 UTC. If a user wants to add the timezone for a lat/lon area that will be up to them. Will clean it up. Nice catch. Thank you.

FoamyGuy (approved):
Looks good to me. Thank you @DJDevon3!

Issue #128, opened 2023-03-02
Please re-work examples to not use deprecated `set_socket`


jepler:

hauserkristof:
Hey @jepler !

Can you help me please, that what is the current correct usage instead of `set_socket` ?

Thank you!

jepler:
Roughly, instead of
```
import adafruit_requests as requests
requests.set_socket(...)
requests.get(...)
```
the new structure is
```
import adafruit_requests
requests = adafruit_requests.Session(...)
requests.get(...)
```
however, I'm not sure that this will actually work for esp32spi-based boards ("airlift"). I've asked @FoamyGuy via discord about it.

toxicantidote:
For anyone who finds this while Googling how to use this library like I did, check out [this Adafruit tutorial](https://learn.adafruit.com/pico-w-wifi-with-circuitpython/pico-w-requests-test-adafruit-quotes) for a more specific, working example.

justmobilize:
I am working through a refactor for requests and once that is done, will be updating and testing all the examples

justmobilize:
All examples in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/151 have been updated to use `adafruit_requests.Session` and `os.getenv`

justmobilize:
@jepler https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/151 is merged. Are the changes sufficient to close this?

anecdata:
I think jepler is out. Closing due to merge of #151.

PR #146, opened 2023-12-17
Move socket specific code from `requests.Session` to `requests.SocketHandler`

justmobilize:
This comes from a thought and discussion to make sockets more usable and have fewer errors when connecting to multiple services (`HTTP`, `MQTT`, etc.).

I have come up with the following proposed steps:

1. Move socket specific code from `requests.Session` to `requests.SocketHandler`
   - This is the Draft PR
2. Make `requests.SocketHandler` a `singleton`
3. Add some methods for better `key<>socket` handling, and helpers like `get_open_keys()`
4. Add tests for new methods and changes
5. Make sure Docs look good
6. At this point mark this PR as `Ready for review`
7. Update `adafruit_esp32spi_socket.py` to `adafruit_esp32spi_socketpool.py`
   - Since `adafruit_esp32spi_socket.py` returns a fake `Socketpool` it would be easier to follow code with more correct naming
8. Update `Adafruit_CircuitPython_MiniMQTT` to take SocketHandler in `__init__` and use it
9. Add something like `SocketHandler.lock_socket()`
   - Which would make sure that you could always open a socket for a particular key: for example `("io.adafruit.com", 1883, "mqtt:")` so you can always connect to AdafruitIO
10. More update thoughts to come

As a side-note, I would **love** to put `SocketHandler` in it's own library, but feel that would break so many examples and things people have made...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ _build
 .idea
 .vscode
 *~
+
+# tox-specific files
+.tox
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -176,7 +176,7 @@ class Response:
         http = self._readto(b" ")
         if not http:
             if session:
-                session._close_socket(self.socket)
+                session._socket_handler.close_socket(self.socket)
             else:
                 self.socket.close()
             raise RuntimeError("Unable to read HTTP response.")
@@ -320,7 +320,8 @@ class Response:
                     self._throw_away(chunk_size + 2)
                 self._parse_headers()
         if self._session:
-            self._session._free_socket(self.socket)  # pylint: disable=protected-access
+            # pylint: disable=protected-access
+            self._session._socket_handler.free_socket(self.socket)
         else:
             self.socket.close()
         self.socket = None
@@ -429,8 +430,8 @@ class Response:
         self.close()
 
 
-class Session:
-    """HTTP session that shares sockets and ssl context."""
+class SocketHandler:
+    """Socket handler for sharing sockets and ssl context."""
 
     def __init__(
         self,
@@ -442,14 +443,24 @@ class Session:
         # Hang onto open sockets so that we can reuse them.
         self._open_sockets = {}
         self._socket_free = {}
-        self._last_response = None
 
-    def _free_socket(self, socket: SocketType) -> None:
+    def _free_sockets(self) -> None:
+        free_sockets = []
+        for sock, val in self._socket_free.items():
+            if val:
+                free_sockets.append(sock)
+
+        for sock in free_sockets:
+            self.close_socket(sock)
+
+    def free_socket(self, socket: SocketType) -> None:
+        """Mark a socket as free so it can be reused if needed"""
         if socket not in self._open_sockets.values():
             raise RuntimeError("Socket not from session")
         self._socket_free[socket] = True
 
-    def _close_socket(self, sock: SocketType) -> None:
+    def close_socket(self, sock: SocketType) -> None:
+        """Close a socket"""
         sock.close()
         del self._socket_free[sock]
         key = None
@@ -460,17 +471,16 @@ class Session:
         if key:
             del self._open_sockets[key]
 
-    def _free_sockets(self) -> None:
-        free_sockets = []
-        for sock, val in self._socket_free.items():
-            if val:
-                free_sockets.append(sock)
-        for sock in free_sockets:
-            self._close_socket(sock)
-
-    def _get_socket(
-        self, host: str, port: int, proto: str, *, timeout: float = 1
+    def get_socket(
+        self,
+        host: str,
+        port: int,
+        proto: str,
+        *,
+        timeout: float = 1,
+        is_ssl: bool = False
     ) -> CircuitPythonSocketType:
+        """Get socket and connect"""
         # pylint: disable=too-many-branches
         key = (host, port, proto)
         if key in self._open_sockets:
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -46,3 +46,6 @@ _build
 .idea
 .vscode
 *~
+
```
justmobilize, line 49: Added as I ran tests locally with tox

adafruit_requests.py
```diff
@@ -429,8 +430,8 @@ def iter_content(self, chunk_size: int = 1, decode_unicode: bool = False) -> byt
         self.close()
 
 
-class Session:
-    """HTTP session that shares sockets and ssl context."""
+class SocketHandler:
```
justmobilize, line 433: Took all the `socket` based methods from `Session` and moved them to `SocketHandler`

adafruit_requests.py
```diff
@@ -442,14 +443,24 @@ def __init__(
         # Hang onto open sockets so that we can reuse them.
         self._open_sockets = {}
         self._socket_free = {}
-        self._last_response = None
 
-    def _free_socket(self, socket: SocketType) -> None:
+    def _free_sockets(self) -> None:
```
justmobilize, line 447: This method stayed the same, just moved up as it's now the only pseudo private one

adafruit_requests.py
```diff
@@ -442,14 +443,24 @@ def __init__(
...
+        free_sockets = []
+        for sock, val in self._socket_free.items():
+            if val:
+                free_sockets.append(sock)
+
+        for sock in free_sockets:
+            self.close_socket(sock)
+
+    def free_socket(self, socket: SocketType) -> None:
```
justmobilize, line 456: `free_socket`, `close_socket`, and `get_socket` became regular methods

adafruit_requests.py
```diff
@@ -460,25 +471,26 @@ def _close_socket(self, sock: SocketType) -> None:
...
-    def _get_socket(
-        self, host: str, port: int, proto: str, *, timeout: float = 1
+    def get_socket(
+        self,
+        host: str,
+        port: int,
+        proto: str,
+        *,
+        timeout: float = 1,
+        is_ssl: bool = False
```
justmobilize, line 481: added `is_ssl` for things like `MQTT`. Example:
```
socket = socket_handler.get_socket("io.adafruit.com", 1883, "mqtt:", is_ssl=True)
```

adafruit_requests.py
```diff
@@ -762,17 +789,33 @@ def wrap_socket(
         return _FakeSSLSocket(socket, self._iface.TLS_MODE)
 
 
+def create_fake_ssl_context(
```
justmobilize, line 792: needed to break out `create_fake_ssl_context` to beable to pre-create a `SocketHandler` when using and `AirLift` or such via `adafruit_esp32spi`

adafruit_requests.py
```diff
@@ -762,17 +789,33 @@ def wrap_socket(
...
+) -> _FakeSSLContext:
+    """Legacy API for creating a fake SSL context"""
+    if not iface:
+        # pylint: disable=protected-access
+        iface = pool._the_interface
+    pool.set_interface(iface)
+    return _FakeSSLContext(iface)
+
+
+def set_socketpool(
```
justmobilize, line 803: In my opinion `set_socket` isn't the right name, since we are actually setting a fake `Socketpool`

justmobilize:
Here is the current coverage report:
image

justmobilize:
I am also happy to tackle https://github.com/adafruit/Adafruit_CircuitPython_Requests/issues/128 and move the handling of `_FakeSSLSocket` into `SocketHandler` for when `is_ssl is True` and `self._ssl_context is None`

adafruit_requests.py
justmobilize: Also, with `create_fake_ssl_context` you can do:
```
ssl_context = adafruit_requests.create_fake_ssl_context(pool, esp)
requests = adafruit_requests.Session(pool, ssl_context, socket_handler)
```
_with or without adding `socket_handler`_

So could completely remove `set_socket` and the new `set_socketpool`

justmobilize:
Tested with a `M4 Feather Express + AirLift FeatherWing` using:
```
import adafruit_esp32spi.adafruit_esp32spi_socket as pool

esp = adafruit_esp32spi.ESP_SPIcontrol(spi, esp32_cs, esp32_ready, esp32_reset)
ssl_context = adafruit_requests.create_fake_ssl_context(pool, esp)

socket_handler = adafruit_requests.SocketHandler(pool, ssl_context)
requests = adafruit_requests.Session(pool, ssl_context, socket_handler)
```

And a `ESP32-S2 via MagTag`
```
ssl_context = ssl.create_default_context()

socket_handler = adafruit_requests.SocketHandler(pool, ssl_context)
requests = adafruit_requests.Session(pool, ssl_context, socket_handler)
```

jepler:
I don't know for sure if anyone's doing this, but what if two ssl contexts are needed? It seems possible that you could need two ssl contexts, such as one with the normal set of certificates, and one with a special certificate needed for a single host that has a non-standard certificate.

Currently, this could be handled by two different Session objects. Looking at how the SocketHandler is used in the examples above, I see that the `ssl_context` is repeated in two places so I'm not sure how that would go.

justmobilize:
That is True. Either you could create 2 different `SocketHandlers` (if they stay not a singleton), or the public `get_socket_handler()` could take in `(pool, ssl_context)` and return the once already created for that pair.

I think you would run into the same issue as current with 2 sessions and the fact they would stomp on top of each other and you would run out of sockets quickly.

justmobilize:
Following up, the `Session` can totally keep track of the `ssl_context`, and only in the case it's not passed in would it generate one...

So from:
```
socket = self._socket_handler.get_socket(host, port, proto, timeout=timeout)
```

to:
```
socket = self._socket_handler.get_socket(host, port, proto, timeout=timeout, ssl_context=self._ssl_context)
```

This way, `SocketHandler` can safely be a singleton and not need to worry about handling multiple contexts

Issue #148, opened 2023-12-28
Issue with sending header values twice


justmobilize:
When helping a member in the discord channel `#help-with-circuitpython`, I had made the mistake of thinking that this library didn't automatically deal with turning a `dict` into a `urlencoded` string. I had them convert it by hand and try and it worked.

As it turns out - it does, but also sends the header `Content-Type: application/x-www-form-urlencoded`, which some systems (`Spotify` as one) can't handle getting the header twice.

I would like to open a PR to check the header for an existing `Content-Type` before adding it's own (both for `urlencoded` and `json`). It would implement a method to check for an existing value since we would also want to check for casing mis-matches.

brentru:
@justmobilize  I am re-opening this issue as https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149 is causing a "TypeError: wrong number of arguments" error as seen here: https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/issues/115. I am also able to locally reproduce this.


The issue appears to be with doing a GET with only headers like in `adafruit_io.py`'s `get` function:

```
    def _get(self, path: str):
        """
        GET data from Adafruit IO

        :param str path: Formatted Adafruit IO URL from _compose_path
        """
        with self._http.get(
            path, headers=self._create_headers(self._aio_headers[1])
        ) as response:
            self._handle_error(response)
            json_data = response.json()
        return json_data
```

Tracing through this, your new `send_header` expects (`self, socket, header, value)`. However, in this case, the header is the `X-AIO-KEY` and the `value` is of `None` type.


On L561, in `send_header`, the `value` is sent via `_send_as_bytes` but it is a `None`Type where `send_as_bytes` expects a `String`:
`self._send_as_bytes(socket, value)`

@justmobilize  Would you be able to resolve this? Do you want me to resolve this?

justmobilize:
@brentru I'm happy to take a look at this

justmobilize:
@brentru looking, my PR just changed from `.encode` to `bytes`

Old:
<img width="335" alt="image" src="https://github.com/adafruit/Adafruit_CircuitPython_Requests/assets/1547881/149d7a40-6713-4088-9a57-bc075d8bf7e6">

Looking at `CPython` requests:
[...]

Issue #40, opened 2020-10-09
Breaking change in 1.7.0: Header keys are now capitalized


makermelissa:
I noticed between 1.6.0 and 1.7.0, headers such as headers['content-length'] are not headers['Content-Length'], so anything that uses the head keys will need to be updated.

This issue is more of a breaking change alert due to the removal of this line:
```
title = str(title.lower(), "utf-8")
```
in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/31/

I'm not sure if we want to put it back or just slowly change everything that broke.

tannewt:
Looks like regular requests is case-insensitive: https://requests.readthedocs.io/en/master/user/quickstart/#response-headers

Maybe we need a wrapper object to do it that way. I removed the title.lower() because it allocates a second copy of the string when one already exists.

makermelissa:
We could have added .lower() onto the title as it is put into the dict. To handle the differences in the PyPortal, I submitted a PR that creates a new dict and just copies the headers with lower case. This way both older and newer versions of this library should work with it.

makermelissa:
By wrapper object, are you suggesting adding a function like get_header() that would accept lowercase and uppercase or maybe a property that lowercases it on the fly or something else?

dhalbert:
Regular `requests` has a `CaseInsensitiveDict` class for exactly this reason:
https://github.com/psf/requests/blob/3bb13f8fbb9d4ed1a20bd33495cdc087eb062ca0/requests/structures.py#L37

tannewt:
> We could have added .lower() onto the title as it is put into the dict. To handle the differences in the PyPortal, I submitted a PR that creates a new dict and just copies the headers with lower case. This way both older and newer versions of this library should work with it.

Calling `.lower()` is copying the string. I'm really trying to avoid that.

> By wrapper object, are you suggesting adding a function like get_header() that would accept lowercase and uppercase or maybe a property that lowercases it on the fly or something else?

No, I'm thinking along the lines of what Dan pointed out. Create an object that acts like a dictionary and does a case insensitive search on each access. That way you only pay the cost on the headers you care about.

makermelissa:
Ok cool. Thanks. I just wanted to make sure I understood.

justmobilize:
Is this still valid? Some header fixes were recently merged in, and if this still exists, I'm happy to work on it.

justmobilize:
[...]

PR #149, opened 2023-12-28
Fix duplicate header issue

justmobilize:
When helping a member in the discord channel `#help-with-circuitpython`, I had made the mistake of thinking that this library didn't automatically deal with turning a `dict` into a `urlencoded` string. I had them convert it by hand and try and it worked.

As it turns out - it does, but also sends the header `Content-Type: application/x-www-form-urlencoded`, which some systems (`Spotify` as one) can't handle getting the header twice.

This would fix issue https://github.com/adafruit/Adafruit_CircuitPython_Requests/issues/148

Test:
```
from os import getenv
import adafruit_requests
import socketpool
import ssl
import wifi

wifi_ssid = getenv("CIRCUITPY_WIFI_SSID")
wifi_password = getenv("CIRCUITPY_WIFI_PASSWORD")
client_id = getenv("SPOTIFY_CLIENT_ID")
client_secret = getenv("SPOTIFY_CLIENT_SECRET")

ssl_context=ssl.create_default_context()
wifi.radio.connect(wifi_ssid, wifi_password)

requests = adafruit_requests.Session(pool, ssl_context)

data = {"grant_type": "client_credentials", "client_id": client_id, "client_secret": client_secret}
headers = {"Content-Type": "application/x-www-form-urlencoded"}
url = "https://accounts.spotify.com/api/token"
r = requests.post(url, data=data, headers=headers)
print(r.status_code)
print(r.text)
```
Using the main branch of `Adafruit_CircuitPython_Requests`, this will return a `400` because it sends `Content-Type` twice, using this branch will succeed

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -530,6 +530,13 @@ class Session:
         return sock
 
     @staticmethod
+    def _header_suppplied(header, supplied_headers):
+        for supplied_header in supplied_headers:
+            if supplied_header.lower() == header.lower():
+                return True
+        return False
+
+    @staticmethod
     def _send(socket: SocketType, data: bytes):
         total_sent = 0
         while total_sent < len(data):
@@ -551,6 +558,16 @@ class Session:
                 raise OSError(errno.EIO)
             total_sent += sent
 
+    def _send_as_bytes(self, socket: SocketType, data: str):
+        return self._send(socket, bytes(data, "utf-8"))
+
+    def _send_header(self, socket, header, value):
+        self._send_as_bytes(socket, header)
+        self._send_as_bytes(socket, ": ")
+        self._send_as_bytes(socket, value)
+        self._send_as_bytes(socket, "\r\n")
+
+    # pylint: disable=too-many-arguments
     def _send_request(
         self,
         socket: SocketType,
@@ -561,40 +578,48 @@ class Session:
         data: Any,
         json: Any,
     ):
-        # pylint: disable=too-many-arguments
-        self._send(socket, bytes(method, "utf-8"))
-        self._send(socket, b" /")
-        self._send(socket, bytes(path, "utf-8"))
-        self._send(socket, b" HTTP/1.1\r\n")
-        if "Host" not in headers:
-            self._send(socket, b"Host: ")
-            self._send(socket, bytes(host, "utf-8"))
-            self._send(socket, b"\r\n")
-        if "User-Agent" not in headers:
-            self._send(socket, b"User-Agent: Adafruit CircuitPython\r\n")
-        # Iterate over keys to avoid tuple alloc
-        for k in headers:
-            self._send(socket, k.encode())
-            self._send(socket, b": ")
-            self._send(socket, headers[k].encode())
-            self._send(socket, b"\r\n")
+        # Convert data
+        content_type_header = None
+
+        # If json is sent, set content type header and convert to string
         if json is not None:
             assert data is None
+            content_type_header = "application/json"
             data = json_module.dumps(json)
-            self._send(socket, b"Content-Type: application/json\r\n")
-        if data:
-            if isinstance(data, dict):
-                self._send(
-                    socket, b"Content-Type: application/x-www-form-urlencoded\r\n"
-                )
-                _post_data = ""
-                for k in data:
-                    _post_data = "{}&{}={}".format(_post_data, k, data[k])
-                data = _post_data[1:]
-            if isinstance(data, str):
-                data = bytes(data, "utf-8")
-            self._send(socket, b"Content-Length: %d\r\n" % len(data))
+
+        # If data is sent and it's a dict, set content type header and convert to string
+        if data and isinstance(data, dict):
+            content_type_header = "application/x-www-form-urlencoded"
+            _post_data = ""
+            for k in data:
+                _post_data = "{}&{}={}".format(_post_data, k, data[k])
+            # remove first "&" from concatenation
+            data = _post_data[1:]
+
+        # Convert str data to bytes
+        if data and isinstance(data, str):
+            data = bytes(data, "utf-8")
+
+        self._send_as_bytes(socket, method)
+        self._send_as_bytes(socket, " /")
+        self._send_as_bytes(socket, path)
+        self._send_as_bytes(socket, " HTTP/1.1\r\n")
+
+        # Send headers
+        if not self._header_suppplied("Host", headers):
+            self._send_header(socket, "Host", host)
+        if not self._header_suppplied("User-Agent", headers):
+            self._send_header(socket, "User-Agent", "Adafruit CircuitPython")
+        if content_type_header and not self._header_suppplied("Content-Type", headers):
+            self._send_header(socket, "Content-Type", content_type_header)
+        if data and not self._header_suppplied("Content-Length", headers):
+            self._send_header(socket, "Content-Length", str(len(data)))
+        # Iterate over keys to avoid tuple alloc
+        for header in headers:
+            self._send_header(socket, header, headers[header])
         self._send(socket, b"\r\n")
+
+        # Send data
         if data:
             self._send(socket, bytes(data))
 
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -530,13 +530,6 @@ class Session:
         return sock
 
     @staticmethod
-    def _header_suppplied(header, supplied_headers):
-        for supplied_header in supplied_headers:
-            if supplied_header.lower() == header.lower():
-                return True
-        return False
-
-    @staticmethod
     def _send(socket: SocketType, data: bytes):
         total_sent = 0
         while total_sent < len(data):
@@ -563,9 +556,9 @@ class Session:
 
     def _send_header(self, socket, header, value):
         self._send_as_bytes(socket, header)
-        self._send_as_bytes(socket, ": ")
+        self._send(socket, b": ")
         self._send_as_bytes(socket, value)
-        self._send_as_bytes(socket, "\r\n")
+        self._send(socket, b"\r\n")
 
     # pylint: disable=too-many-arguments
     def _send_request(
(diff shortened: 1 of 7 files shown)


adafruit_requests.py
```diff
@@ -529,6 +529,13 @@ def _get_socket(
         self._socket_free[sock] = False
         return sock
 
+    @staticmethod
+    def _header_suppplied(header, supplied_headers):
```
justmobilize, line 533: Added method to verify if a header exists already

adafruit_requests.py
```diff
@@ -551,6 +558,16 @@ def _send(socket: SocketType, data: bytes):
                 raise OSError(errno.EIO)
             total_sent += sent
 
+    def _send_as_bytes(self, socket: SocketType, data: str):
+        return self._send(socket, bytes(data, "utf-8"))
+
+    def _send_header(self, socket, header, value):
```
justmobilize, line 564: Removed duplicate code with a single send header method

adafruit_requests.py
```diff
@@ -551,6 +558,16 @@ def _send(socket: SocketType, data: bytes):
                 raise OSError(errno.EIO)
             total_sent += sent
 
+    def _send_as_bytes(self, socket: SocketType, data: str):
```
justmobilize, line 561: simplify code with a send as bytes method

adafruit_requests.py
```diff
@@ -561,40 +578,48 @@ def _send_request(
...
-            self._send(socket, b"\r\n")
-        if "User-Agent" not in headers:
-            self._send(socket, b"User-Agent: Adafruit CircuitPython\r\n")
-        # Iterate over keys to avoid tuple alloc
-        for k in headers:
-            self._send(socket, k.encode())
-            self._send(socket, b": ")
-            self._send(socket, headers[k].encode())
-            self._send(socket, b"\r\n")
+        # Convert data
```
justmobilize, line 581: Updated method to
1. modify data and determine possible `Content-Type` header
2. send init bytes
3. send `host` and `user agent` headers (when not supplied)
4. send `content type` and `content length` headers (when not supplied)
5. send supplied headers 
6. send data

tests/header_test.py
```diff
@@ -30,7 +98,55 @@ def _send(data):
     requests_session.get("http://" + HOST + "/get", headers=headers)
 
     sock.connect.assert_called_once_with((IP, 80))
-    sent = b"".join(sent).lower()
+    sent = b"".join(sent)
     assert b"user-agent: blinka/1.0.0\r\n" in sent
-    # The current implementation sends two user agents. Fix it, and uncomment below.
```
justmobilize, line 35: Fixed this bug

tests/header_test.py
```diff
@@ -12,7 +12,75 @@
 RESPONSE_HEADERS = b"HTTP/1.0 200 OK\r\nContent-Length: 0\r\n\r\n"
 
 
-def test_json():
+def test_host():
```
justmobilize, line 15: added tests to verify `host`, `user agent` and `content type` are sent

tests/post_test.py
```diff
@@ -64,10 +65,18 @@ def test_form():
     pool.socket.return_value = sock
 
     requests_session = adafruit_requests.Session(pool)
-    data = {"Date": "July 25, 2019"}
+    data = {"Date": "July 25, 2019", "Time": "12:00"}
```
justmobilize, line 68: added more data to verify the `&` is included

tests/post_test.py
```diff
@@ -64,10 +65,18 @@ def test_form():
...
     requests_session.post("http://" + HOST + "/post", data=data)
     sock.connect.assert_called_once_with((IP, 80))
-    sock.send.assert_called_with(b"Date=July 25, 2019")
+    sock.send.assert_has_calls(
```
justmobilize, line 71: test that the `content type` header was automatically added

tests/post_test.py
```diff
@@ -77,7 +86,15 @@ def test_json():
     pool.socket.return_value = sock
 
     requests_session = adafruit_requests.Session(pool)
-    json_data = {"Date": "July 25, 2019"}
+    json_data = {"Date": "July 25, 2019", "Time": "12:00"}
     requests_session.post("http://" + HOST + "/post", json=json_data)
     sock.connect.assert_called_once_with((IP, 80))
-    sock.send.assert_called_with(b'{"Date": "July 25, 2019"}')
+    sock.send.assert_has_calls(
```
justmobilize, line 92: test that the `content type` header was automatically added

justmobilize:
This may also resolve: https://github.com/adafruit/Adafruit_CircuitPython_Requests/issues/40

adafruit_requests.py
```diff
@@ -529,6 +529,13 @@ def _get_socket(
         self._socket_free[sock] = False
         return sock
 
+    @staticmethod
+    def _header_suppplied(header, supplied_headers):
+        for supplied_header in supplied_headers:
+            if supplied_header.lower() == header.lower():
+                return True
+        return False
```
dhalbert, line 537: This seems kind of laborious to me, and `.lower()`s each header key multiple times. You could put the `.lower()` version of all the header keys in a `set()`, and then just check set membership.

adafruit_requests.py
```diff
@@ -551,6 +558,16 @@ def _send(socket: SocketType, data: bytes):
                 raise OSError(errno.EIO)
             total_sent += sent
 
+    def _send_as_bytes(self, socket: SocketType, data: str):
+        return self._send(socket, bytes(data, "utf-8"))
+
+    def _send_header(self, socket, header, value):
+        self._send_as_bytes(socket, header)
+        self._send_as_bytes(socket, ": ")
```
dhalbert, line 566: Cheaper, because no run-time call to `bytes()`.
```suggestion
        self._send(socket, b": ")
```

adafruit_requests.py
```diff
@@ -551,6 +558,16 @@ def _send(socket: SocketType, data: bytes):
...
+        self._send_as_bytes(socket, value)
+        self._send_as_bytes(socket, "\r\n")
```
dhalbert, line 568:
```suggestion
        self._send_as_bytes(socket, b"\r\n")
```

adafruit_requests.py
```diff
@@ -561,40 +578,48 @@ def _send_request(
...
+            data = _post_data[1:]
+
+        # Convert str data to bytes
+        if data and isinstance(data, str):
+            data = bytes(data, "utf-8")
+
+        self._send_as_bytes(socket, method)
+        self._send_as_bytes(socket, " /")
+        self._send_as_bytes(socket, path)
+        self._send_as_bytes(socket, " HTTP/1.1\r\n")
```
dhalbert, line 606:
```suggestion
        self._send(socket, b" HTTP/1.1\r\n")
```

adafruit_requests.py
```diff
@@ -561,40 +578,48 @@ def _send_request(
...
+                _post_data = "{}&{}={}".format(_post_data, k, data[k])
+            # remove first "&" from concatenation
+            data = _post_data[1:]
+
+        # Convert str data to bytes
+        if data and isinstance(data, str):
+            data = bytes(data, "utf-8")
+
+        self._send_as_bytes(socket, method)
+        self._send_as_bytes(socket, " /")
```
dhalbert, line 604:
```suggestion
        self._send(socket, b" /")
```

dhalbert (changes requested):
Hi - several suggestions.

adafruit_requests.py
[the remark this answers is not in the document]
justmobilize: typed too fast... fixed in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149/commits/5fa5aa218bbaecdd7b81c1e33a0894de635a8eff

adafruit_requests.py
justmobilize: fixed in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149/commits/5fa5aa218bbaecdd7b81c1e33a0894de635a8eff

adafruit_requests.py
justmobilize: fixed in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149/commits/5fa5aa218bbaecdd7b81c1e33a0894de635a8eff

adafruit_requests.py
justmobilize: fixed in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149/commits/5fa5aa218bbaecdd7b81c1e33a0894de635a8eff

adafruit_requests.py
justmobilize: fixed in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149/commits/5fa5aa218bbaecdd7b81c1e33a0894de635a8eff

adafruit_requests.py
justmobilize: That's a great idea (used to making simple methods in CPython). fixed in https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/149/commits/5fa5aa218bbaecdd7b81c1e33a0894de635a8eff

FoamyGuy (approved):
This looks good to me. Thanks for the fix @justmobilize!

I confirmed the duplicated Content-Type header issue with a simple http server and wireshark to view the raw TCP data stream. I tried a few other things before resorting to wireshark, but flask and a few of the other basic server examples I tried made it difficult for me to see the raw request. Many of them treat headers as dictionary-like so it seems to resolve duplicates without much trouble. 

I confirmed the fix does successfully get rid of that duplicated header. 

All testing was was done on a QTPy ESP32-S2 8.2.9

justmobilize:
> This looks good to me. Thanks for the fix @justmobilize!
> 
> I confirmed the duplicated Content-Type header issue with a simple http server and wireshark to view the raw TCP data stream. I tried a few other things before resorting to wireshark, but flask and a few of the other basic server examples I tried made it difficult for me to see the raw request. Many of them treat headers as dictionary-like so it seems to resolve duplicates without much trouble.
> 
> I confirmed the fix does successfully get rid of that duplicated header.
> 
> All testing was was done on a QTPy ESP32-S2 8.2.9

Thanks for testing! And happy to do this (and I hope a bunch more)

FoamyGuy:
Looks like all requested changes have been addressed. Merging this one. Thanks again for all of your work on this @justmobilize!

PR #151, opened 2024-01-23
Use ConnectionManager

justmobilize:
Update `adafruit_requests` to use [Adafruit_CircuitPython_ConnectionManager](https://github.com/adafruit/Adafruit_CircuitPython_ConnectionManager/) for socket handling. As well as clean up examples


justmobilize:
Updated the RTD example section to have all 4 boards listed:
image

justmobilize:
@tannewt asking here, just so we have record of the conversation.

The code that is currently [here](https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/c533f184d7e1af70a8aac281055f38f555c93641/adafruit_requests.py#L619C1-L680C57) which is re-worked from the old `set_socket` to set a global instance - Is this something we want to keep?

Since `set_socket` is gone, everyone will still need to move to passing in a `socket_pool` and `ssl_context` and thus aren't making anything easier by leaving this here.

justmobilize:
@tannewt I'm also moving this to `ready for review`. I would like to add some more tests before it merges one we know what's what

tannewt:
> @tannewt asking here, just so we have record of the conversation.
> 
> The code that is currently [here](https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/c533f184d7e1af70a8aac281055f38f555c93641/adafruit_requests.py#L619C1-L680C57) which is re-worked from the old `set_socket` to set a global instance - Is this something we want to keep?
> 
> Since `set_socket` is gone, everyone will still need to move to passing in a `socket_pool` and `ssl_context` and thus aren't making anything easier by leaving this here.

Ya, let's just remove it in favor of Session().

examples/wifi/expanded/requests_wifi_api_github.py
```diff
@@ -22,11 +20,12 @@
...
 sleep_time = 900
 
-try:
-    from secrets import secrets
-except ImportError:
-    print("Secrets File Import Error")
-    raise
+# Get WiFi details, ensure these are setup in settings.toml
+ssid = os.getenv("CIRCUITPY_WIFI_SSID")
+appw = os.getenv("CIRCUITPY_WIFI_PASSWORD")
```
tannewt, line 25: What is `appw` for?

justmobilize:
> > @tannewt asking here, just so we have record of the conversation.
> > The code that is currently [here](https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/c533f184d7e1af70a8aac281055f38f555c93641/adafruit_requests.py#L619C1-L680C57) which is re-worked from the old `set_socket` to set a global instance - Is this something we want to keep?
> > Since `set_socket` is gone, everyone will still need to move to passing in a `socket_pool` and `ssl_context` and thus aren't making anything easier by leaving this here.
> 
> Ya, let's just remove it in favor of Session().

@tannewt after removing and adding tests, any other changes?

examples/wifi/expanded/requests_wifi_api_github.py
justmobilize: I just standardized in what there was the most of. Assumed it was an internal preference. Can change them all to `password` or anything else that makes sense.

justmobilize: Ahhh- `AP PassWord`

tannewt: I'd prefer `password`. This is the only other change from me.

justmobilize:
@tannewt this is ready for review again. Coverage is at 85%, I tackled what was easy and added some fixtures to remove boilerplate. Will add more tests in a future pass.

image

tannewt (approved):
Looks great! Thank you!

PR #152, opened 2024-01-26
Correctly handle headers of different types

justmobilize:
This stems from [this issue](https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/issues/115) in `Adafruit_CircuitPython_AdafruitIO` that caused [this issue](https://github.com/adafruit/Adafruit_CircuitPython_Requests/issues/148) to be re-opened.

If this path looks good I will add tests for the changes.

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -530,6 +530,18 @@ class Session:
         return sock
 
     @staticmethod
+    def _check_headers(headers: Dict[str, str]):
+        if not isinstance(headers, dict):
+            raise AttributeError("headers must be in dict format")
+
+        for key, value in headers.items():
+            if isinstance(value, (str, bytes)) or value is None:
+                continue
+            raise AttributeError(
+                f"Header part ({value}) from {key} must be of type str or bytes, not {type(value)}"
+            )
+
+    @staticmethod
     def _send(socket: SocketType, data: bytes):
         total_sent = 0
         while total_sent < len(data):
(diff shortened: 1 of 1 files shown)


adafruit_requests.py
```diff
@@ -529,6 +529,18 @@ def _get_socket(
...
+    def _check_headers(headers: Dict[str, str]):
+        if not isinstance(headers, dict):
+            raise AttributeError("headers must be in dict format")
+
+        for key, value in headers.items():
+            if isinstance(value, (str, bytes)) or value is None:
+                continue
+            raise AttributeError(
+                f"Header part ({value}) from {key} must be of type str or bytes, not {type(value)}"
+            )
```
justmobilize, line 542: in `CPython requests` [this is](https://github.com/psf/requests/blob/96b22fa18c00831656ee4b286bf1c9062459b00a/src/requests/utils.py#L1040):
```
        raise InvalidHeader(
            f"Invalid leading whitespace, reserved character(s), or return "
            f"character(s) in header {header_kind}: {header_part!r}"
        )
```
An `AttributeError` seems similar enough and didn't know if creating a new Exception would be wanted

adafruit_requests.py
```diff
@@ -555,9 +567,14 @@ def _send_as_bytes(self, socket: SocketType, data: str):
         return self._send(socket, bytes(data, "utf-8"))
 
     def _send_header(self, socket, header, value):
+        if value is None:
```
justmobilize, line 570: In `CPython requests` they have a method [merge_setting](https://github.com/psf/requests/blob/96b22fa18c00831656ee4b286bf1c9062459b00a/src/requests/sessions.py#L61)

That they use in a bunch of places, but it creates new dict objects, and thought this would be simpler in memory usage

FoamyGuy (approved):
This looks good to me. Agree with both of the points you mentioned in other comments. 

Thanks for the fix @justmobilize!

I tested this successfully with a Feather ESP32S3 TFT with:
- requests simpletest
- adafruit io simpletest
- adafruit io http simpletest

All appear to function as intended with this version.

brentru:
@FoamyGuy  Can we release this library, is it picked up by Adabot?

Issue #138, opened 2023-07-07
Poor exception handling (lack of meaningful messages) for low available memory (below ~41k)


jake1164:
I have run into issues with using the Requests library on a Pico W  while getting JSON responses from an https url. The exception messages will vary based on the amount of free memory. I have created a quick [sample ](https://github.com/jake1164/requests_low_memory_example) to test and show the behavior. 

The sample just loads images into a list and prints out the JSON returned until the board runs low on memory.  

Below ~41k the exceptions make it look like a certification error and has the message 'Sending request failed'
Below ~27k it just fails and gives no exception message at all.  

The same result happens with `requests.get(url)` and `requests.get(url, stream=True)`. 

Crudely tested on: 
CircuitPython 8.2.0, 8.1.0, 8.0.5, 8.0.2
Requests 1.14.1, 1.13.4, 1.12.9
```
before 13488
appended images 44160
appended image 44000
Getting json
getting url: https://api.weather.gov/alerts/active?area={state}
free memory 44000
free memory after 6224
{'correlationId': '606fcf6', 'status': 400, 'instance': 'https://api.weather.gov/requests/606fcf6', 'detail': 'Bad Request', 'type': 'https://api.weather.gov/problems/BadRequest', 'title': 'Bad Request', 'parameterErrors': [{'parameter': 'query.area[0]', 'message': 'Does not have a value in the enumeration ["AM","AN","GM","LC","LE","LH","LM","LO","LS","PH","PK","PM","PS","PZ","SL"]'}, {'parameter': 'query.area[0]', 'message': 'Does not have a value in the enumeration ["AL","AK","AS","AR","AZ","CA","CO","CT","DE","DC","FL","GA","GU","HI","ID","IL","IN","IA","KS","KY","LA","ME","MD","MA","MI","MN","MS","MO","MT","NE","NV","NH","NJ","NM","NY","NC","ND","OH","OK","OR","PA","PR","RI","SC","SD","TN","TX","UT","VT","VI","VA","WA","WV","WI","WY","MP","PW","FM","MH"]'}, {'parameter': 'query.area[0]', 'message': 'Failed to match exactly one schema'}]}
before 11184
appended images 41856
appended image 41696
Getting json
getting url: https://api.weather.gov/alerts/active?area={state}
free memory 41696
x509-crt-bundle:PK verify failed with error FFFFBD70
x509-crt-bundle:Failed to verify certificate
requests.get(url): Sending request failed
{}
[...]

Issue #130, opened 2023-03-29
It's unfortunate that Response.close() tries to read all remaining information; buggy too


jepler:
Response.close() seems to try to read all remaining data out of the socket in order to work around a flaw or limitation in the ESP32SPI socket implementation.

There are two problems with this:
 * A client may want / need to close an ongoing request before reading all information in it
 * as implemented it's not necessarily exhausting all content, but at most one chunk in the chunked-encoding
   * this may not be possible, the stream could be infinite
 * handling of chunked mode is wrong and can lead to errors during close()

As far as the last point, I did experience it 'going wrong' with the chunked mode length being part of the http stream, but there's also the case where the exact content is carefully read (runs on desktop linux):
```py
import socket
import ssl
import adafruit_requests

session = adafruit_requests.Session(socket, ssl.create_default_context())
with session.get('https://github.com') as response:
    assert response._chunked
    buf = bytearray(1)
    response._readinto(buf)
    print(buf)
    while response._remaining:
        print(response._remaining)
        response._readinto(buf)
        print(buf)
    assert response._cached is None
    assert response.socket
```

if there's a limitation of ESP32SPI that a socket has to be fully read before it's closed, can't we push that down into ESP32SPI (or better yet fix the limitation in nina-fw) and get rid of this trouble-causing block?

justmobilize:
@jepler or @dhalbert can this be closed with my cleanup on `response.close()`

dhalbert:
Fixed by #159.

PR #156, opened 2024-02-28
close(): fix exception and enable fast-close

bablokb:
The current code throws an exception during `close()` when it tries to convert an empty buffer (`chunk_header`).
This PR also adds an optional parameter `fast=False`. Passing `True` will skip reading.

This addresses #130 (also mentioned in #138)

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -226,22 +226,23 @@ class Response:
         while to_read > 0:
             to_read -= self._recv_into(buf, to_read)
 
-    def close(self) -> None:
+    def close(self, fast: bool = False) -> None:
         """Drain the remaining ESP socket buffers. We assume we already got what we wanted."""
         if not self.socket:
             return
         # Make sure we've read all of our response.
-        if self._cached is None:
+        if self._cached is None and not fast:
             if self._remaining and self._remaining > 0:
                 self._throw_away(self._remaining)
             elif self._chunked:
                 while True:
                     chunk_header = bytes(self._readto(b"\r\n")).split(b";", 1)[0]
+                    if not chunk_header:
+                        break
                     chunk_size = int(bytes(chunk_header), 16)
                     if chunk_size == 0:
                         break
                     self._throw_away(chunk_size + 2)
-                self._parse_headers()
         if self._session:
             # pylint: disable=protected-access
             self._session._connection_manager.free_socket(self.socket)
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -86,7 +86,9 @@ class Response:
 
     encoding = None
 
-    def __init__(self, sock: SocketType, session: "Session", fast_close: bool = False) -> None:
+    def __init__(
+        self, sock: SocketType, session: "Session", fast_close: bool = False
+    ) -> None:
         self.socket = sock
         self.encoding = "utf-8"
         self._cached = None
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -363,7 +363,7 @@ class Session:
         socket_pool: SocketpoolModuleType,
         ssl_context: Optional[SSLContextType] = None,
         session_id: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


tannewt:
Let's not add it to `close()` because that changes it from the Requests API: https://requests.readthedocs.io/en/latest/api/#requests.Response.close

Instead, please add it as a kwarg on `Session()`. The `Session()` API already varies from CPython.

tannewt (changes requested):
Please move kwarg to `fast_close` kwarg in `Session()`.

bablokb:
Maybe I am missing something, but this would imply that I

- add a kwarg to Session()
- the constructor saves this as a public property
- `Response.__init__()` will query the property from the Session-object passed to it and save it as an instance attribute
- `Response.close()` uses this instance attribute

This seems rather involved. In my solution, I added an optional argument to `Response.close()`, so as long as you don't use it, the code is compatible. And the whole `Response` class is not compatible after all, there are many differences already.

But probably the whole drain-the-buffer-code should not be here after all, but somewhere where the socket is actually closed. Until this is implemented, I see both of my changes as a workaround. Instead of changing the code in many places, it might be wiser to just remove the additional option and wait for the real solution.

tannewt:
> And the whole `Response` class is not compatible after all, there are many differences already.

What are these differences? The intent is that the same code will work with CPython's request.

> This seems rather involved.

It is a bit of plumbing but it isn't complex *and* it preserves the API. I'd have session pass it into the response object as a separate kwarg.

> In my solution, I added an optional argument to `Response.close()`, so as long as you don't use it, the code is compatible.

"so as long as you don't use it" is a big caveat. We shouldn't make it possible to write incompatible code so you don't need to worry about it or need to redo it later.

bablokb:
> What are these differences? The intent is that the same code will work with CPython's request.

e.g. the constructor of `Response` in CPython does not take any arguments.

> I'd have session pass it into the response object as a separate kwarg.

What is the fundamental difference between adding an additional kwarg to the `Response` constructor compared to adding a kwarg to `Response.close()`?

Anyhow, I will do what you suggested to get the fix for the chunk_size exception merged, that is the most important problem.

tannewt:
> > What are these differences? The intent is that the same code will work with CPython's request.
> 
> e.g. the constructor of `Response` in CPython does not take any arguments.
> 
> > I'd have session pass it into the response object as a separate kwarg.
> 
> What is the fundamental difference between adding an additional kwarg to the `Response` constructor compared to adding a kwarg to `Response.close()`?

Response objects aren't usually created by user code. Instead, they are constructed by the library internally and returned. To me, that makes the constructor more private than close.

> 
> Anyhow, I will do what you suggested to get the fix for the chunk_size exception merged, that is the most important problem.

Thanks!

justmobilize:
@bablokb what is your thought around `fast_close`? I understand the `chunk` bug fix, but don't fully understand this one.

dhalbert:
Is there a CPython `requests` equivalent to the `fast_close` functionality, or is there something different about our `requests` semantics?

bablokb:
> @bablokb what is your thought around `fast_close`? I understand the `chunk` bug fix, but don't fully understand this one.

There are streams that never end (e.g. internet radio mp3-streams). In this case reading the socket until it is empty will never succeed.

dhalbert:
I can revert the PR if it makes sense, or if you can just rebase and fix, @justmobilize, that's fine too.

justmobilize:
@dhalbert I should have my PR ready by Wednesday. I can do either:
1. include the `empty_chucks` and fix and not having close keep reading read
2. remove the `fast_close` and fix and not having close keep reading read
Up to you on what's best (revert now or wait)

dhalbert:
@justmobilize Do 2., so that we have at least one fix for now.

justmobilize:
Sounds good

bablokb:
Tested it with 9.0.0-rc0 and a Pico-W and it works fine.

Issue #153, opened 2024-02-15
`requests.head` doesn't work


jepler:
The following code fails with etimedout (wifi setup not shown):
```python
r = requests.head("http://wifitest.adafruit.com/testwifi/index.html", timeout=3)
r.close()
```

```
Traceback (most recent call last):
  File "code.py", line 41, in <module>
  File "adafruit_requests.py", line 313, in close
  File "adafruit_requests.py", line 301, in _throw_away
  File "adafruit_requests.py", line 203, in _recv_into
OSError: [Errno 116] ETIMEDOUT
```

This appears to occur because a HEAD request may contain a content-length header, but it never actually contains any content.

A quick and dirty workaround:
```diff
diff --git a/adafruit_requests.py b/adafruit_requests.py
index 26b40f0..76c1ec4 100644
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -160,7 +160,12 @@ class Response:
 
     encoding = None
 
-    def __init__(self, sock: SocketType, session: Optional["Session"] = None) -> None:
+    def __init__(
+        self,
+        sock: SocketType,
+        session: Optional["Session"] = None,
+        is_head: bool = False,
+    ) -> None:
         self.socket = sock
         self.encoding = "utf-8"
         self._cached = None
@@ -185,6 +190,8 @@ class Response:
         self.reason: bytearray = self._readto(b"\r\n")
         """The status reason returned by the server"""
         self._parse_headers()
+        if is_head:
+            self._remaining = 0
         self._raw = None
         self._session = session
 
@@ -695,7 +703,7 @@ class Session:
         if not socket:
             raise OutOfRetries("Repeated socket failures") from last_exc
 
-        resp = Response(socket, self)  # our response
+        resp = Response(socket, self, method == "HEAD")  # our response
         if allow_redirects:
             if "location" in resp.headers and 300 <= resp.status_code <= 399:
                 # a naive handler for redirects
```

but tagging @justmobilize for interest

justmobilize:
@jepler since I was also going to look at https://github.com/adafruit/Adafruit_CircuitPython_Requests/issues/130. Which is the actual problem here. I recommend fixing that one (and looking at `ESP32SPI`) which would fix this one. Thoughts?

justmobilize:
@jepler or @dhalbert can this be closed with my cleanup on `response.close()`

dhalbert:
Sure: fixed by #156 and #159.

[...]

PR #159, opened 2024-03-12
Don't read all on response.close()

justmobilize:
I tested on a `M4 Express + Airlift` and a `UM FeatherS3`. I tried every combination of switching hosts, partial reads, etc to see if I could get an error without reading the whole response and didn't get any errors

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -86,14 +86,11 @@ class Response:
 
     encoding = None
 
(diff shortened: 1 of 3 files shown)


justmobilize:
Should close #130 and #153

dhalbert (approved):
Thanks for this and for the testing!

PR #167, opened 2024-03-18
Update Open-Sky Network Private Single Flight API Example

DJDevon3:
Requires Open-Sky Network website account. Instructions in code comments. Allows for more daily calls than public API.

Added
- pretty serial hierarchy view
- general code cleanup and improvements
- improved endpoints with server side failure messages
- improved and streamlined debugging
- shouty constants to make Pylint happy.
- imports sorted
- switched from circuitpython_base64 to built-in module binascii for base64 credential encoding
- doc strings for functions
- uses settings.toml by default
- adafruit_circuitpython_requests now using Connection Manager by default

Serial output example:
```
Connecting to WiFi...
✅ Wifi!
 | Attempting to GET OpenSky-Network Single Private Flight JSON!
 | Website Credentials Required! Allows more daily calls than Public.
 | ✅ OpenSky-Network JSON!
 |  | Last Contact: 03/19/2024 09:49:10
 |  | Transponder: 4b1806
 |  | Callsign: SWR30M  
 |  | Squawk: 1000
 |  | Origin: Switzerland
 |  | Longitude: 8.7047
 |  | Latitude: 48.1189
 |  | Barometric Altitude: 3962.4
 |  | Velocity: 176.06
 |  | Vertical Rate: 0
✂️ Disconnected from OpenSky-Network API

Finished!
Board Uptime: 1.2 days
Next Update: 30 minutes
===============================

### CODE REVIEW

**Comment by DJDevon3** [comment]
I messed up and published both the private and public API examples here. 
commit 8a34d2a is the private example
commit 231e2eb is the public example
The commits didn't merge because I published them from 2 branches. sigh, I don't know how to correct this.

**Comment by dhalbert** [comment]
> The commits didn't merge because I published them from 2 branches. sigh, I don't know how to correct this.

You can revert a commit. Or fix it locally and do a `push -f` (force push).

**Comment by FoamyGuy** [comment]
I think I am in favor of keeping them separate.  I wouldn't say that it makes a huge difference either way, but I like that the seperate public API one can be a bit less complex (losing the chunk of code responsible for base 64 encoding the auth), as well as drop the requirement for `adafruit_binascii`

I often start my projects by copying the most relavent library example for the first feature that I decide to work on, I appreciate having the seperated minimal examples because it tends to result in less unused stuff to find and cut out of the code.

**Comment by FoamyGuy** [comment]
@DJDevon3 I'm don't think I understand your comment about the two different commits for public / private versions of the examples. 

Is that something that needs to get resolved before this PR can be considered for merging? 

Maybe your intention was originally to do public and private in separate PRs? but they accidentally got both put into this one? If that is the case I think it's okay to just move forward with them both in this PR.

The changes look mostly fine to me, I'll go ahead and test them out but will hold off on approval / merging just in case there is something that needs to be done prior to that.

File: examples/wifi/expanded/requests_wifi_api_openskynetwork_private.py [hunk 1]
```diff
@@ -19,135 +18,177 @@
...
+pool = adafruit_connection_manager.get_radio_socketpool(wifi.radio)
+ssl_context = adafruit_connection_manager.get_radio_ssl_context(wifi.radio)
+requests = adafruit_requests.Session(pool, ssl_context)
+
+# -- Base64 Conversion --
+OSN_CREDENTIALS = str(osnusername) + ":" + str(osnpassword)
+OSN_CREDENTIALS_B = b"" + str(OSN_CREDENTIALS) + ""
+BASE64_ASCII = b2a_base64(OSN_CREDENTIALS_B)
+BASE64_STRING = str(BASE64_ASCII)  # bytearray
+TRUNCATED_BASE64_STRING = BASE64_STRING[2:-1]  # truncate bytearray head/tail
```
FoamyGuy, line 49: Is this line's intention to cut out the `b'` at the beginning and `'` at the end of a bytestring representation to basically "convert" it to a regular string?

It seems like that is what it's doing to me, if so I think there are better ways to convert between string and bytestring without having to replace those "manually" like this. If my understanding of the intention is correct I'll figure out the specific syntax for the other ways to make that conversion.

FoamyGuy (commented):
I tested the public example successfully. But I'm having trouble getting the private API example to work. I double checked that my username and password are correct by logging in via the website with the same ones and using the debug flag to have it print them locally in my serial. 

It seems like the Base64 version is ending up with an extra `\n` at the end of the base64 data, I'm not sure if that is intentional or if it's related to the problem for certain but it's odd because it isn't there in the password.

examples/wifi/expanded/requests_wifi_api_openskynetwork_private.py
DJDevon3: Yes that was the intention for the base64 string to just be a string because it needs to be inserted into the POST as a concatenated url. I think there's a way to use .decode possibly but manually slicing works. I tested each one and each iteration prior to submitting.  I'll have to take a look into why it's not slicing properly.

The private api's require website login and provide a higher daily call limit. The public API's do not require a login or token.  I might not have named the files the best names for this.  I just made them up as I went.  If you think there's a better naming convention for these API examples please feel free to rename them. 

Yes I intended to keep private and public as separate branches and accidentally committed 2 PRs from separate branches to the same branch. :(  It's definitely a problem that needs to be resolved. I can probably do it by deleting the branches and remaking them unless you have a better way.  

I do not have a code backup so please don't delete them.  I've had 3 hours of sleep and running on empty.  I can attempt to resolve tomorrow.

DJDevon3:
### About "\n" appended to base64 hash:
I switched this and many other examples that relied on base64 encoding from the community library circuitpython_base64 to adafruit_binascii as suggested by Dan.  I thought the "\n" on the end was part of the hash.  Oops! Obviously I didn't do enough testing.  Open-Skynetwork accepts it with the \n appended with or without the slice! for me anyway, and that's quite odd.  

This isn't a real login, just an example base64 hash example
```
VXNlcm5hbWU6MDEyMzQ1Njc4OQ==
VXNlcm5hbWU6MDEyMzQ1Njc4OQ==\n
```
Open-Sky Network will accept either of these for me... and that's not good.  I'm glad it fails for you. It should fail for me but it doesn't.

### Issue with binascii library:
The binacii simpletest doesn't include a way to spit out the hashed string without b'encapsulation\n'.  The library is adding the \n... WHY?!?!  It's even demonstrated in the bin_ascii simpletest.

If they would have provided an example to drill it down to a pure string I'm sure they would have also used slice[2:-3] method or a function to output to a pure string.  I don't see a method to do that.  This is a problem for passing the exact correct string into the header. As far as I can tell it's presumed that the developer (you and me) must use slice to get it the rest of the way. I have a lot of branch splitting and recommits to do today anyway, will definitely fix this.

FoamyGuy:
Interesting about binascii adding the newline. It appears that behavior is inherieted from CPython: https://docs.python.org/3/library/binascii.html#binascii.b2a_base64 although in CPYthon the eventually added a `newline` argument to b2a_base64 in order to disable the newline being added https://github.com/python/cpython/issues/69544

I think we should add same newline argument and behavior into adafruit_binascii so it can match the CPython implementation. 

I don't necessarily think that adafruit_binascii would be the appropriate place for a function that takes username:password and spits out the string ready to be added to a header unless the CPYthon version has that functionality. For the circuitpython libraries that match CPython named ones they should be a subset of the CPython one with matching behavior whenever possible. I do think the functionality itself would be valuable but should get added somewhere that matches a CPython API if possible. 

In CPython requests it looks like there is `HTTPBasicAuth` which handles that https://requests.readthedocs.io/en/latest/user/authentication/ we could add that class here in the adafruit_requests library and then have matching APIs between circuitpython and CPython more closely which will make code more portable.

FoamyGuy:
Actually it looks like there is also a core module `binascii` and it does already have the newline argument.

```
>>> import binascii
>>> dir(binascii)
['__class__', '__name__', '__dict__', 'a2b_base64', 'b2a_base64', 'crc32', 'hexlify', 'unhexlify']
>>> binascii.b2a_base64(b"hello")
b'aGVsbG8=\n'
>>> binascii.b2a_base64(b"hello", newline=False)
b'aGVsbG8='
```
I ran the above on a M5 Cardputer (ESP32 S3 based)

So it looks like you could use this and won't even need to have `adafruit_binascii` as a required library if I am understanding correctly (assuming this module is enabled in the core on all or most devices, I only tried 1 so far)

DJDevon3:
I did see that last night when looking into the library but was unsure how to use it or if it was a usable parameter. Thank you for the example.  I'll take a look into updating them after the dust settles from all the PR's.

DJDevon3:
Figured out why I'm still able to login even with \n appended. It only checks the first 20 characters of the hash, ignores anything after the 20th character lol.  So \n should have also worked for you unless your hash is less than 20 characters.

```
VXNlcm5hbWU6MDEyMzQ1Njc4OQ==  # this will log you in
VXNlcm5hbWU6MDEyMzQ  #  so will this
```
However if the base64 has ends up being shorter than 20 characters it will check them all... I hope! I haven't tested it since I can't change my username to be shorter.

DJDevon3:
Also used this commit to delete the twitter api example and use ruff with pre-commit as an experiment.

examples/wifi/expanded/requests_wifi_api_openskynetwork_public.py
```diff
@@ -17,121 +17,164 @@
...
-transponder = "ab1644"
+TRANSPONDER = "3c5ef8"
 
-# Initialize WiFi Pool (There can be only 1 pool & top of script)
-pool = socketpool.SocketPool(wifi.radio)
+# Get WiFi details, ensure these are setup in settings.toml
+ssid = os.getenv("CIRCUITPY_WIFI_SSID")
+password = os.getenv("CIRCUITPY_WIFI_PASSWORD")
+osnusername = os.getenv("OSN_USERNAME")  # Website Credentials
+osnpassword = os.getenv("OSN_PASSWORD")  # Website Credentials
```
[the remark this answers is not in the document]
DJDevon3: yep, missed that.  removed that in the public single flight example, nice catch. thank you!

DJDevon3:
This one is ready to go.  Please merge this one first, then [PR 172](https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/172)

FoamyGuy:
> Figured out why I'm still able to login even with \n appended. It only checks the first 18 characters of the hash, ignores anything after 

Yikes! :fearful:

examples/wifi/expanded/requests_wifi_api_openskynetwork_private.py
```diff
@@ -19,135 +18,178 @@
...
+# -- Base64 Conversion --
+OSN_CREDENTIALS = str(osnusername) + ":" + str(osnpassword)
+# base64 encode and strip appended \n from bytearray
+OSN_CREDENTIALS_B = binascii.b2a_base64(b"" + str(OSN_CREDENTIALS)).strip()
+BASE64_STRING = str(OSN_CREDENTIALS_B)  # bytearray
+SLICED_BASE64_STRING = BASE64_STRING[2:-1]  # slice bytearray head/tail
+
+if DEBUG:
+    print("Base64 ByteArray: ", BASE64_STRING)
+    print(f"Base64 Sliced String: {SLICED_BASE64_STRING}")
```
FoamyGuy, line 53:
```suggestion
OSN_CREDENTIALS_B = binascii.b2a_base64(OSN_CREDENTIALS.encode()).strip()
BASE64_STRING = OSN_CREDENTIALS_B.decode()  # bytearray

if DEBUG:
    print("Base64 String: ", BASE64_STRING)

```

examples/wifi/expanded/requests_wifi_api_openskynetwork_private.py
```diff
@@ -19,135 +18,178 @@
...
 
 # Requests URL - icao24 is their endpoint required for a transponder
 # example https://opensky-network.org/api/states/all?icao24=a808c5
-# OSN private requires your username:password to be base64 encoded
-osn_header = {"Authorization": "Basic " + str(base64cred)}
-OPENSKY_SOURCE = "https://opensky-network.org/api/states/all?" + "icao24=" + transponder
+# OSN private: requires your website username:password to be base64 encoded
+OPENSKY_HEADER = {"Authorization": "Basic " + str(SLICED_BASE64_STRING)}
```
FoamyGuy, line 58:
```suggestion
OPENSKY_HEADER = {"Authorization": "Basic " + BASE64_STRING}
```

FoamyGuy (changes requested):
lets use `encode()` and `decode()` instead of the slicing to convert from bytestring to string. 

I've tested the changes suggested bellow successfully and they appear to be working successfully. 

I think this is the proper way to do those conversions and it allows us to get rid of `SLICED_BASE64_STRING` which makes that portion of the code a little bit less complex.

examples/wifi/expanded/requests_wifi_api_openskynetwork_private.py
DJDevon3: Very nice.  That makes things easier.  Didn't know how to use encode and decode.  Learning a lot here thank you!  

I'm not sure what's up with their hash checking.  Maybe an attempt at optimization, maybe lazy, or a bit of both.  Either way it's not the right way to do it, I'm tempted to try to contact them and request they check the full hash.  Odds are if you have the first 18 of the hash the rest is going to be correct but it's still not the right way to do it.

DJDevon3:
I'm using pylint ignore module import due to this.
```py
C:\Users\Devon\Documents\GitHub\Adafruit_CircuitPython_Requests\examples\wifi\expanded>black requests_wifi_api_openskynetwork_private.py
All done! ✨ 🍰 ✨
1 file left unchanged.

C:\Users\Devon\Documents\GitHub\Adafruit_CircuitPython_Requests\examples\wifi\expanded>pylint requests_wifi_api_openskynetwork_private.py
************* Module requests_wifi_api_openskynetwork_private
requests_wifi_api_openskynetwork_private.py:10:0: E0401: Unable to import 'adafruit_connection_manager' (import-error)
requests_wifi_api_openskynetwork_private.py:11:0: E0401: Unable to import 'wifi' (import-error)

-------------------------------------------------------------------
Your code has been rated at 9.12/10 (previous run: 10.00/10, -0.88)
``` This is on my local pylint.  It has helped me clean up a lot of errors before submitting but some I cannot get passed.

Will try to submit even with the error and see what happens.

DJDevon3:
Apparently that passes here but not on my local pylint.  Makes me wonder how many other linting errors I've been fixing unnecessarily.

FoamyGuy (approved):
Looks good. Thank you!

PR #171, opened 2024-03-18
Switch to ruff

justmobilize:


.pre-commit-config.yaml
```diff
@@ -3,45 +3,19 @@
...
-          - --disable=consider-using-f-string,duplicate-code,missing-docstring,invalid-name
-      - id: pylint
-        name: pylint (test code)
-        description: Run pylint rules on "tests/*.py" files
-        types: [python]
-        files: "^tests/"
-        args:
-          - --disable=consider-using-f-string,duplicate-code,missing-docstring,invalid-name,protected-access,redefined-outer-name
+      - id: ruff
+        args: ["--select", "I", "--fix"]
```
tekktrik, line 16: `--fix` is a good one since it will help with local development, but it might help to add the rules to `pyproject.toml`:  (example: https://github.com/tekktrik/circfirm/blob/ab296f8a33db1ceab0b0e5f1fbeb1dc89923396c/pyproject.toml#L82-L84)

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: 2024 Justin Myers for Adafruit Industries
+#
+# SPDX-License-Identifier: Unlicense
+
+* text eol=lf
```
tekktrik, line 5: Is the `.gitattributes` for helping with Windows?  I haven't had an issue with `ruff` and EOLs.

.pre-commit-config.yaml
```diff
@@ -3,45 +3,19 @@
...
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: trailing-whitespace
-  - repo: https://github.com/pycqa/pylint
-    rev: v2.17.4
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.1.3
```
FoamyGuy, line 13: This appears to automatically redirect to `https://github.com/astral-sh/ruff-pre-commit` I assume Github will keep that redirect active, but it's kind of an undocumented oddity for anyone that runs across it looking for the current name. I would recommend changing here to the current URL instead of relying on the redirect. 

I noticed they are also up to release `0.3.2`, is there any reason why we would need to use 0.1.3 instead of the newer / current release?

FoamyGuy:
I created a similar draft over here in Display_Text library: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/pull/204 to get my head wrapped around the changes and see another example of what kinds of stuff it might be wanting to change. 

The vast majority of changes in display text were tweaks to import order and extra whitespace lines sprinkled it breaking up imports and comments. 

I've not seen anything in the code that looks like it would be problematic on CircuitPython but I do still intend to test a few of these libraries post ruff on a device.

.gitattributes
FoamyGuy: It is for helping with windows, not specifically related to `ruff` but was discussed in the same larger conversation during the meeting.

Though my understanding is this was a fix for something that pylint complained about, perhaps with ruff replacing pylint we may not need this change any more?

.pre-commit-config.yaml
justmobilize: All for what ever rules make sense. Was trying to keep things simple for the example

.gitattributes
justmobilize: This will make sure they commit fine, even if they forget to run pre-commit, allowing small changes to still pass CI

.pre-commit-config.yaml
justmobilize: Based on comments, I grabbed what MicroPython was using. Totally forgot to mention that.

justmobilize:
> I created a similar draft over here in Display_Text library: [adafruit/Adafruit_CircuitPython_Display_Text#204](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/pull/204) to get my head wrapped around the changes and see another example of what kinds of stuff it might be wanting to change.
> 
> The vast majority of changes in display text were tweaks to import order and extra whitespace lines sprinkled it breaking up imports and comments.
> 
> I've not seen anything in the code that looks like it would be problematic on CircuitPython but I do still intend to test a few of these libraries post ruff on a device.

Also saw some trailing comma changes, which I believe you can turn off (although I like them, because future changes are more clear)

FoamyGuy:
Not strictly related to ruff, but something else that would be nice to check for during the same automation is confirming the docs conf.py setting for `autoclass_content = 'both'` There is at least one library adafruit_hid that was found to differ with this configuration. It would be convenient to find and fix any others with the same sweep. More context on discord here: https://discord.com/channels/327254708534116352/327298996332658690/1219668121468669962

PR #179, opened 2024-04-01
Remove pylint disable

FoamyGuy:
It seems like these are unnecessary. I didn't notice it in some of the other recent updates so these ones already were merged with it. 

This PR removes the ones currently merged.


justmobilize:
Makes sense to me. If we switch to linting tools these all go away then too

PR #189, opened 2024-04-21
Files arg

FoamyGuy:
This change adds a `files` argument that can be passed to `post()` 

The value supported is dictionary with string keys and tuple values. The behavior when passed matches the behavior of CPython requests. 

A new example was added illustrating the functionality, I've only tested with native wifi on ESP32 S3, so that is the only one added.

I was able to confirm the upload is working correctly by using a local simple flask server that acceps file upload and by examining the result returned by httpbin.org when the new example runs. It returns the image back in base64 format. This cyberchef recipe was used to validate that the original image does pop back out the other side https://cyberchef.org/#recipe=From_Base64('A-Za-z0-9%2B/%3D',true,false)Render_Image('Raw')

Note that the CPYthon implementation of this also supports a `files` dictionary with string keys and file_handle values (as opposed to tupple containing the file_handle and other things). That form of the argument is not supported on circuitpython. In order for that to be supported I believe we would need access to FileIO.name property or something that provides similar functionality. I've created [#9195](https://github.com/adafruit/circuitpython/issues/9195) in the core inquiring about adding that, but at present time this is not supported in the circuitpython version.


adafruit_requests.py
```diff
@@ -441,13 +451,70 @@ def _send_request(
...
+
+        # pylint: disable=too-many-nested-blocks
+        if files is not None and isinstance(files, dict):
+            boundary_str = _generate_boundary_str()
+            content_type_header = f"multipart/form-data; boundary={boundary_str}"
+
+            for fieldname in files.keys():
+                if not fieldname.endswith("-name"):
+                    if files[fieldname][0] is not None:
+                        file_content = files[fieldname][1].read()
```
justmobilize, line 474: @FoamyGuy I have something similar in progress, but chunked out the read during the send, so it allows for sending large files. What are your thoughts?

FoamyGuy: @justmobilize is that functionality that is supported by CPython requests too? Is it the same as this? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked_encoding or am I misunderstanding that? 

Generally I'd say I'm in favor of any functionality that also exists in CPython unless it happens to require substantially large code size for marginal benefits.

I don't have any real experience with the chunked transfers, and the main project that I had in mind for using this functionality is uploading relatively small files so it wasn't something I attempted to implement or looked into but that doesn't mean I'm opposed to it.

justmobilize: So it's more just internal to how we do it. So instead of building one huge data object that goes to:
```py
self._send(socket, bytes(data))
```
We read the files there in chunks and then send. So we're never using more than 100 (or whatever makes sense) bytes at a time.

FoamyGuy: It sounds like a good improvement to me, but I don't have a clear idea how the code would look to make that work without interrupting the non-file upload requests and with requests that inlcude multiple "parts" where some of them are plain form values like `"othervalue": (None, "HelloWorld"),` from the example. 

If you've got an idea of how that would work or even have it started already I'd say feel free to add commit(s) to this branch to adapt it to work like that, or if you'd prefer you can fork this and link me to your fork with those changes and I can merge it back to here. Or it could always be a follow up PR. I don't have any strong preference, whatever is easiest for you is good with me.

justmobilize: I'm working on some other PRs this afternoon. I'll fork this and move some things around and share it

justmobilize: @FoamyGuy merged what you have and what I had and drafted a PR here: https://github.com/FoamyGuy/Adafruit_CircuitPython_Requests/pull/1

Added some comments of things it still needs

Let me know your thoughts

dhalbert:
@FoamyGuy is this ready for final review or is it still in progress/

FoamyGuy:
I believe this is ready for review now. I've merged Justin's changes that were submitted against my branch to this so they are included here now.

justmobilize:
@dhalbert I created a new PR into his branch to add some tests back in. Would love to see those in (can add later if wanted)

justmobilize:
@dhalbert my tests are merged and this is ready for review!

adafruit_requests.py
```diff
@@ -357,10 +360,74 @@ def __init__(
...
+            content_length += len(boundary_data)
+            boundary_objects.append(boundary_data)
+
+            if hasattr(file_handle, "read"):
+                is_binary = False
+                try:
+                    content = file_handle.read(1)
+                    is_binary = isinstance(content, bytes)
+                except UnicodeError:
+                    is_binary = False
```
dhalbert, line 395: We could really use `file_handle.mode` here, hunh? No other way to find out whether the file is opened as text or binary.

adafruit_requests.py
```diff
@@ -357,10 +360,74 @@ def __init__(
...
+
+        return boundary_string, content_length, boundary_objects
+
+    @staticmethod
+    def _build_boundary_string():
+        hex_characters = "0123456789abcdef"
+        _boundary = ""
+        for _ in range(32):
+            _boundary += random.choice(hex_characters)
+        return _boundary
```
dhalbert, line 425: how about `os.urandom(16).hex()` instead?

adafruit_requests.py
```diff
@@ -357,10 +360,74 @@ def __init__(
...
+                boundary_data += f'; filename="{file_name}"'
+            boundary_data += "\r\n"
+            if len(field_values) >= 3:
+                file_content_type = field_values[2]
+                boundary_data += f"Content-Type: {file_content_type}\r\n"
+            if len(field_values) >= 4:
+                file_headers = field_values[3]
+                for file_header_key, file_header_value in file_headers.items():
+                    boundary_data += f"{file_header_key}: {file_header_value}\r\n"
+            boundary_data += "\r\n"
```
dhalbert, line 384: You might want to try to reduce the amount of string concatenation, because that generates garbage and potentially causes fragmentation. For instance, lines 372 and 373 could be one string. You could collect all the strings in a list and then `join` them at the end.

dhalbert (changes requested):
A few suggestions.

adafruit_requests.py
justmobilize: Since the concatenated string goes into `boundary_objects`, would it be best just to put the smaller strings in there?

adafruit_requests.py
justmobilize: I didn't remember seeing a `mode`. Should it exist?

adafruit_requests.py
dhalbert: I wasn't sure whether it mattered how boundary objects was structured. I thought it might be a mix of strings and bytes. But if it's a bunch of chunks that are just sent out, and the structure doesn't matter, then the "objects" could be fragments.

adafruit_requests.py
dhalbert: It exists in CPython! Not in MicroPython/CircuitPython. I was just remarking here's a case where it would come in handy. Maybe it would be easy and small to add.

justmobilize:
@FoamyGuy shall I make another PR into this branch with the changes?

adafruit_requests.py
FoamyGuy: Nice, I didn't know about that one. Changed to that with the latest commit.

adafruit_requests.py
FoamyGuy: Initially I attempted to use multi-line string with triple quotes, but I found that it was causing trouble with string to bytestring conversions and switched to the current structure with lots of concatenating strings.

FoamyGuy:
> @FoamyGuy shall I make another PR into this branch with the changes?

I got the change to os.urandom() with the latest commit, if you've got a clear idea how to tackle reducing the string concatenation then definitely feel free to PR to this branch. 

I'll try to take a crack at it over the weekend if you don't have the opportunity to get it first.

justmobilize:
@FoamyGuy created here: https://github.com/FoamyGuy/Adafruit_CircuitPython_Requests/pull/4

adafruit_requests.py
FoamyGuy: This was changed to use the list instead of string concatenation with the latest commits merged in from Justin's PR to my branch.

justmobilize:
@dhalbert ready to go again!

adafruit_requests.py
```diff
@@ -357,10 +360,66 @@ def __init__(
...
+            if hasattr(file_handle, "read"):
+                is_binary = False
+                try:
+                    content = file_handle.read(1)
+                    is_binary = isinstance(content, bytes)
+                except UnicodeError:
+                    is_binary = False
+
+                if not is_binary:
+                    raise AttributeError("Files must be opened in binary mode")
```
dhalbert, line 398: ValueError is for problems with attributes, but not for problems with values of attributes.
See https://docs.python.org/3/library/exceptions.html#ValueError
https://docs.python.org/3/library/exceptions.html#AttributeError
```suggestion
                    raise ValueError("Files must be opened in binary mode")
```

adafruit_requests.py
```diff
@@ -357,10 +360,66 @@ def __init__(
...
+
+    @staticmethod
+    def _build_boundary_string():
+        return os.urandom(16).hex()
+
     @staticmethod
     def _check_headers(headers: Dict[str, str]):
         if not isinstance(headers, dict):
-            raise AttributeError("headers must be in dict format")
+            raise AttributeError("Headers must be in dict format")
```
dhalbert, line 422:
```suggestion
            raise ValueError("Headers must be in dict format")
```

dhalbert (changes requested):
There are more `AttributeError` usages that what I marked. They are in older code that isn't touched by this PR. Could you change those as well? Thanks. The tests will also need to be changed to use `ValueError`.

I looked at the bundle and there are many libraries that use `AttributeError` in places where should use `Value` Error. Maybe I should bring this up in the Monday meeting. There are also some references in Learn code.

This will be a major version change due to the exception changes.

dhalbert:
Maybe the `ValueError` -> `AttributeError` should be a separate PR.

justmobilize:
@dhalbert I'm happy to work through this repo and fix up the exceptions and tests, and can do so after this is merged if that is the only concern.

dhalbert (approved):
I am willing to add this -- I am a little worried about library size but we'll see if it causes issues when frozen.

justmobilize:
@dhalbert I had this idea of a `requests-plus`. That requires normal `requests` but has extra methods and overrides. The other option is that `requests-plus` could just have add-in methods (like in this case `_build_boundary_string`, `_build_boundary_data` etc. And if you pass in `files` and we got an import error for `requests-plus` it would error...

dhalbert:
@FoamyGuy Thanks for merging. This is an upward-compatible improvement, so it deserved a minor version bump. But no need to change the version number now.

PR #193, opened 2024-05-18
Update Exceptions

justmobilize:


justmobilize:
@dhalbert let me know if I missed anything

dhalbert (approved):
Thanks for replacing the `AttributeError`s, and also using `TypeError` where appropriate.

PR #194, opened 2024-06-19
Add ability to post a file via the data param

justmobilize:
Add ability to post a file via the data param
- Broke some common functionality into reusable blocks
- if the `data` object has `read` send it up as raw data, reading in chuncks

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -50,7 +50,7 @@ SEEK_END = 2
 
 if not sys.implementation.name == "circuitpython":
     from types import TracebackType
-    from typing import Any, Dict, Optional, Type
+    from typing import IO, Any, Dict, Optional, Type
 
     from circuitpython_typing.socket import (
         SocketpoolModuleType,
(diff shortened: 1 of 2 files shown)


jmangum:
Tested @justmobilize solution with a Pico W/OV5640 PiCowbell/PiCowbell doubler setup which posts jpeg to Adafruit IO with feed history turned-off.  Works perfectly!  Due to memory limitations have to reset board (I actually put it into a `TimeAlarm` deep sleep) after each picture acquisition and upload, but this is a small price to pay (and probably a more sensible solution for an application that one might want to run with a battery).  Thank you @justmobilize !

jerryneedell:
Is there an example for using this?  I'd like to give it a try, but I'm not sure how to use it.

jmangum:
Sure thing @jerryneedell .  Attached is my test code.  Very much a work-in-progress, so please excuse the awful code.  It does work as I have repeatedly posted pictures to aio.
[code_iotfeeder.py.txt](https://github.com/user-attachments/files/15955501/code_iotfeeder.py.txt)

justmobilize:
@jerryneedell here is the test code I used (just used an existing file): https://gist.github.com/justmobilize/f0ee9668d7aab6d91524b3ab53953d2c

FoamyGuy (approved):
This looks good to me, thanks for the new functionality @justmobilize! 

Thanks for testing and letting us know @jmangum.

I tested this version successfully on a Feather ESP32S3 TFT. I compared requests made with this version of the library against those made with CPython requests using the same requests.post argument values. 

All of my testing was with a local flask test server, not adafruit IO specifically. 


I do think it would be nice ultimately to have an example in the repo in the examples dir, it's slightly more discoverable there than in the tests.

PR #199, opened 2024-07-01
Document socket property of Response objects

### LINKED ISSUE #198: Need supported way to "steal" socket from requests library for mp3 streaming
Right now I do this:
```py

def get_mp3_stream():
    if STREAMING_URL.startswith("http:") or STREAMING_URL.startswith("https:"):
        return requests.get(STREAMING_URL, headers={"connection": "close"}).socket
    return open(STREAMING_URL, "rb")
```
but
 * the socket property is not documented and not in standard `requests`
 * this has no way of properly releasing the socket at the ConnectionManager level

The real requests library has a public `raw` property that calling code may use when the request includes `stream=True`. So simply renaming the `socket` property to `raw` would be one step in the right direction. (the stream argument is already supported in adafruit_requests)

```py
        #: File-like object representation of response (for advanced usage).
        #: Use of ``raw`` requires that ``stream=True`` be set on the request.
        #: This requirement does not apply for use internally to Requests.
        self.raw = None
```

As for properly releasing the socket, I guess I need to restructure my code so it can hold onto the requests object as long as needed.

jepler:
The socket property is a CircuitPython addition not present in "standard requests" (which provides similar but not quite identical functionality via the "raw" property of response objects).

As discussed in the linked issue, having a documented way to access the underlying socket is needed for MP3 streaming over http/https.

Closes #198

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -88,6 +88,21 @@ class Response:
     # pylint: disable=too-many-instance-attributes
 
     encoding = None
+    socket: SocketType
+    """The underlying socket object (CircuitPython extension, not in standard requests)
+
+    Under the following circumstances, calling code may directly access the underlying
+    socket object:
+
+     * The request was made with ``stream=True``
+     * The request headers included ``{'connection': 'close'}``
+     * No methods or properties on the Response object that access the response content
+       may be used
+
+    Methods and properties that access response headers may be accessed.
+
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Thanks for documenting this. As noted, CPython doesn't provide a good way to get the socket -- this is better.

PR #200, opened 2024-07-03
handle servers that omit a space after a colon in response headers

jepler:
Example URL: "http://ice2.somafm.com/dronezone-128-mp3"

```
$ curl -i http://ice2.somafm.com/dronezone-128-mp3
HTTP/1.1 200 OK
Content-Type: audio/mpeg
Date: Wed, 03 Jul 2024 21:17:04 GMT
icy-br:128
...
```

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -251,7 +251,8 @@ class Response:
             header = self._readto(b"\r\n")
             if not header:
                 break
-            title, content = bytes(header).split(b": ", 1)
+            title, content = bytes(header).split(b":", 1)
+            content = content.strip()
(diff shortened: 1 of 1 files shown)


jepler:
If possible please tag a release after merging this example, it's required for an in progress guide.

justmobilize:
Should we add a test, to make sure it doesn't accidentally get reset?

FoamyGuy (approved):
This looks good to me. 

@justmobilize We could add a test for it, but I don't necessarily think we need to have it before merging this and allowing the learn guide to be unblocked.

PR #72, opened 2021-02-06
Catch recv-into exception to prevent Out of sockets

anecdata:
An `ETIMEDOUT` exception occurring on a `socket.recv_into` call in a Session request would leave the socket open, and after two occurrences, would get `RuntimeError: Out of sockets` and require a microcontroller reset.
```
Traceback (most recent call last):
  File "code.py", line 675, in http_get
  File "/lib/adafruit_requests.py", line 595, in get
  File "/lib/adafruit_requests.py", line 572, in request
OSError: [Errno 116] ETIMEDOUT
# ...
Traceback (most recent call last):
  File "code.py", line 675, in http_get
  File "/lib/adafruit_requests.py", line 595, in get
  File "/lib/adafruit_requests.py", line 572, in request
OSError: [Errno 116] ETIMEDOUT
# ...
Traceback (most recent call last):  
  File "code.py", line 675, in http_get
  File "adafruit_requests.py", line 595, in get
  File "adafruit_requests.py", line 559, in request
  File "adafruit_requests.py", line 436, in _get_socket
  File "adafruit_requests.py", line 433, in _get_socket
RuntimeError: Out of sockets
```
After this change, if the failure persists, it will raise an `OutOfRetries: ` exception, just as in the case of persistent `_send_request` failures.

PR #205, opened 2024-09-27
Fixes for exceptions in Session.request leading to sockets not properly closed

aggieNick02:
Fix cases in Session.request method where an exception could lead to exiting the method without the created socket being owned by a response or closed.  This would then cause future calls to get a socket with the same parameters to fail because one already existed and hadn't been closed.  The PR https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/72 fixed this issue for one case where an exception could be raised. This generalizes that in an attempt to fix it in other cases that could still be hit.


tannewt (approved):
Thank you!

Issue #210, opened 2025-05-09
Responses with no content cause timeout error and crash


AlecIsaacson:
I'm using CircuitPython v 9.2.6 on an Adafruit M4 Express Airlift Lite.  The Adafruit_Requests module version 4.0.0 crashes with 116: ETIMEOUT when it receives a response with no content (or alternatively, when it receives a response with HTTP response code 204 - no content).

To reproduce, run the following code, [adapted from the ESP32SPI example found here.](https://docs.circuitpython.org/projects/requests/en/latest/examples.html)

Note how when we get a 200 response, things are fine, but when we get the 204 (with a no content response) we generate a timeout error.  I'd expect adafruit_response to handle no content responses rather than timeout.

```
# SPDX-FileCopyrightText: 2021 ladyada for Adafruit Industries
# SPDX-License-Identifier: MIT

import os

import adafruit_connection_manager
import board
import busio
from adafruit_esp32spi import adafruit_esp32spi
from digitalio import DigitalInOut

import adafruit_requests

# Get WiFi details, ensure these are setup in settings.toml
ssid = os.getenv("CIRCUITPY_WIFI_SSID")
password = os.getenv("CIRCUITPY_WIFI_PASSWORD")

# If you are using a board with pre-defined ESP32 Pins:
esp32_cs = DigitalInOut(board.ESP_CS)
esp32_ready = DigitalInOut(board.ESP_BUSY)
esp32_reset = DigitalInOut(board.ESP_RESET)

spi = busio.SPI(board.SCK, board.MOSI, board.MISO)
radio = adafruit_esp32spi.ESP_SPIcontrol(spi, esp32_cs, esp32_ready, esp32_reset)

print("Connecting to AP...")
while not radio.is_connected:
    try:
        radio.connect_AP(ssid, password)
    except RuntimeError as e:
        print("could not connect to AP, retrying: ", e)
        continue
print("Connected to", str(radio.ap_info.ssid, "utf-8"), "\tRSSI:", radio.ap_info.rssi)

# Initialize a requests session
pool = adafruit_connection_manager.get_radio_socketpool(radio)
ssl_context = adafruit_connection_manager.get_radio_ssl_context(radio)
requests = adafruit_requests.Session(pool, ssl_context)

# The POST URL uses httpbin's standard post interface, which normally returns a 200.
# The 204 URL will cause httpbin to always return a 204.
JSON_POST_URL = "https://httpbin.org/post"
[...]

PR #211, opened 2025-05-11
Dont try to read body of 204 no_content resp

FoamyGuy:
Resolves: #210 

With this change the behavior now matches between CircuitPython and CPython

circuitpython:
```
POSTing data to https://httpbin.org/status/204: 31F
Data received from server: b''
Response code 204
```

CPYthon:
```
POSTing data to https://httpbin.org/status/204: 31F
Data received from server: b''
Response code 204
```

Without this change the code goes into `iter_content()` which gets to readinto and stalls until timeout as there is no data to read.

I am not certain if this is the correct / best way to handle 204 responses, but it does seem to match the behavior of CPython. I tried poking around briefly in the CPYthon requests code to look for any special handling of 204 status like this and couldn't find any so I'm not sure what drives the behavior on CPython.

All of my testing was on ESP32S3 TFT with 9.2.7

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -300,7 +300,10 @@ class Response:
                 return self._cached
             raise RuntimeError("Cannot access content after getting text or json")
 
-        self._cached = b"".join(self.iter_content(chunk_size=32))
+        if self.status_code != 204:
+            self._cached = b"".join(self.iter_content(chunk_size=32))
+        else:
+            self._cached = b""
         return self._cached
 
     @property


justmobilize:
@FoamyGuy I was going to tackle this tomorrow. The Python requests library doesn't do this based on status code, and think we should try to match their logic, thoughts?

anecdata:
200 without content=length and not chunked I thought was fixed by: https://github.com/adafruit/Adafruit_CircuitPython_Requests/pull/123

My understanding (which is shallow) is that CPython can detect when the server closes the socket, using socket.fileno() (or `.select()` or similar. In ESP32SPI I think we have .available or can access he specific states of socket setup and shutdown. But nothing I know of that's comparable in native wifi sockets. Not sure about WIZnet.

204 by definition has no response body:
https://www.rfc-editor.org/rfc/rfc7231.html#section-6.3.5
I may be missing a third way, but it seems we either need to condition on status code, or add capability to sense that the server has closed the socket (supporting ESP32SPI, WIZnet, and two native implementations). Otherwise, that leaves timeout. I don't know if it's valid for the server to `keep-alive` with a 204, that would confound the approach of sensing that the socket is closed.

justmobilize:
A quick look, and that PR uses `_remaining`, which is based on content length, which for a 204 is specifically not returned.

So the correct fix would see that `_remaining` is `None` and probably something else to make sure there wasn't any data...

anecdata:
I'm just wondering if that something needs to be a timeout, or if there's some other mechanism in the absence of content-length, chunked, or socket close. It doesn't seem like curl or CPython Requests have any discernable timeout.

If not conditioning on status code, a 200 is like a 204 except it has no body (a 200 could also have no body, but then it should have a content-length header of 0), but the client may not know there are no body bytes incoming without waiting to see?

justmobilize:
My plan was to step through a normal 204 in CPython requests and see what it gets. And then see if I can reproduce it in a 3 in circuitpython...

anecdata:
I'm glad you all are on it. I'm baffled how it works given all of the above :-)

justmobilize:
@FoamyGuy would like you thoughts. I went down the rabbit-hole and got:

in Cpython 
1. requests: calls [self.raw.read(chunk_size)](https://github.com/psf/requests/blob/c65c780849563c891f35ffc98d3198b71011c012/src/requests/models.py#L832), where `raw` is `urllib3.response.HTTPResponse`
2. urllib3 goes `read`, `_raw_read`, `_fp_read` to [return self._fp.read(amt) if amt is not None else self._fp.read()](https://github.com/urllib3/urllib3/blob/98634f7a5a0f39029a21e9f49e9066a8f99a1515/src/urllib3/response.py#L862), where `_fp` is a `http.client.HTTPResponse`
3. http.client.HTTPResponse [read](https://github.com/python/cpython/blob/f2de1e6861c27bd498f598efc01600450979b5f9/Lib/http/client.py#L494) hits here and `self.fp`is a `_io.BufferedReader`

All this to say that the socket takes care of this. So I see 2 options:
1. Update the core, ESP32SPI and WIZNET5K to handle this correctly
2. Do a hack in requests, like this

I would prefer option 1, because I think we would keep fixing things where the socket code isn't handling things right.

justmobilize:
Looking a little bit more, I see when using `CPython` and this library, it hangs on `_recv_into`, since we have this in `_readinto`:
```
            elif self._remaining is None:
                # the Content-Length is not provided in the HTTP header
                # so try parsing as long as their is data in the socket
                pass
```

which then does:
```
    read = self._recv_into(buf, nbytes)
```

which just sits there...

Everything I read and know, you should always have `Content-Length` unless `Transfer-Encoding` is `chunked`.

I did check a Hue (since #123 is what added that) and it indeed doesn't return a `Content-Length`.

I'm going to dig a little more, to try to find the right fix

justmobilize:
okay, @FoamyGuy finally found it in [http/client.py](https://github.com/python/cpython/blob/9983c7d4416cac8deb2fded1ec9c7daf786c3a02/Lib/http/client.py#L381:L385)
```
        # does the body have a fixed length? (of zero)
        if (status == NO_CONTENT or status == NOT_MODIFIED or
            100 <= status < 200 or      # 1xx codes
            self._method == "HEAD"):
            self.length = 0
```

I think we can move this logic into `Response._parse_headers` to set the length to 0 matching this. So similar to what you have, just a little different and in a different spot

FoamyGuy:
@justmobilize that works for me, do you want to PR that change? we can close this one. I'll submit it in another day or two if you are unable.

PR #214, opened 2025-05-31
use a local test server instead of httpbin

FoamyGuy:
Make our own little test server to run the test against instead of relying on 3rd party httpbin.

### DIFF
--- a/tests/local_test_server.py
+++ b/tests/local_test_server.py
@@ -1,0 +1,38 @@
+# SPDX-FileCopyrightText: 2025 Tim Cocks
+#
+# SPDX-License-Identifier: MIT
+import json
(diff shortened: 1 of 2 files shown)


justmobilize:
@FoamyGuy this looks great, do we also want to add the tests from `files_test.py`? They POST to `https://httpbin.org/post` just to make sure what we send and what CPython requests send are the same

FoamyGuy:
@justmobilize yes that would be good. Ideally I think it's best to use a local server for anything that we can instead of remote httpbin. 

I think it can come in a future PR though, this one doesn't need to wait on it. 

I had a look in `files_test.py` and I'm not sure that I understand how it functions, like what mechanism populates values to the arguments being passed to the tests. I need to read a bit more about pytest.

PR #215, opened 2025-10-03
LICENSE non-exec.

edwbaker:
Having LICENSE executable raises an issue with Lintian when packaging using dh-virtualenv.


dhalbert (approved):
Thanks for spotting this. I was relieved to see that this seems to be the only `LICENSE` file in the bundle that had the executable bits set.

The "Empty file." in the file diff is misleading; it is not empty.

PR #217, opened 2025-11-01
Localize more tests

### LINKED ISSUE #216: Remove Reliance on 3rd party server for tests
one or more of the tests rely on httpbin.org in order to pass. They fail if that server is unresponsive during the request time, here is an example log of one such instance: https://github.com/adafruit/Adafruit_CircuitPython_Requests/actions/runs/18664231889/job/53211702649#step:2:1422

Ideally we should remove reliance on this 3rd party server from the test. Any tests that still rely on httbin should be converted to use: https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/main/tests/local_test_server.py and that simple server script can be expanded as needed to support anything our tests do.

FoamyGuy:
Resolves #216 

Moves all remaining tests that used httpbin.org to use the local test server instead. 

Added two new tests that use `http://www.adafruit.com/api/quotes.php` and `https://www.adafruit.com/api/quotes.php` to test redirects and https.

### DIFF
--- a/tests/real_call_test.py
+++ b/tests/real_call_test.py
@@ -42,11 +42,18 @@ def test_gets():
         adafruit_connection_manager.connection_manager_close_all(release_references=True)
 
 
-def test_http_to_https_redirect():
+@pytest.mark.parametrize(
+    ("allow_redirects", "status_code"),
+    (
+        (True, 200),
+        (False, 301),
+    ),
+)
+def test_http_to_https_redirect(allow_redirects, status_code):
     url = "http://www.adafruit.com/api/quotes.php"
     requests = adafruit_requests.Session(socket, ssl.create_default_context())
-    with requests.get(url) as response:
-        assert response.status_code == 200
+    with requests.get(url, allow_redirects=allow_redirects) as response:
+        assert response.status_code == status_code
 
 
 def test_https_direct():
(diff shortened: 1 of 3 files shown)


tests/real_call_test.py
```diff
@@ -28,25 +29,28 @@ def test_gets():
...
-                if case[json_keys_index] is not None:
-                    for key, value in case[json_keys_index].items():
-                        assert response.json()[key] == value
 
-            adafruit_connection_manager.connection_manager_close_all(release_references=True)
+def test_http_to_https_redirect():
+    url = "http://www.adafruit.com/api/quotes.php"
+    requests = adafruit_requests.Session(socket, ssl.create_default_context())
+    with requests.get(url) as response:
+        assert response.status_code == 200
```
justmobilize: @FoamyGuy worth adding parametrize, to make sure that allow_redirects actually works as well?
```suggestion
@pytest.mark.parametrize(
    ("allow_redirects", "status_code"),
    (
        (True, 200),
        (False, 301),
    ),
)
def test_http_to_https_redirect(allow_redirects, status_code):
    url = "http://www.adafruit.com/api/quotes.php"
    requests = adafruit_requests.Session(socket, ssl.create_default_context())
    with requests.get(url, allow_redirects=allow_redirects) as response:
        assert response.status_code == status_code
```

justmobilize (approved):

PR #218, opened 2026-07-02
add port back to host arguments (sends correct host-header, fixes redirects)

bablokb:
The current version of `request()` strips the  port from the host-part of the  url. Subsequently, `_send_request()` sends a wrong host-header. Also when relative redirects are processed, the missing original port will redirect to a wrong url.

### DIFF
--- a/adafruit_requests.py
+++ b/adafruit_requests.py
@@ -647,7 +647,9 @@ class Session:
             )
             ok = True
             try:
-                self._send_request(socket, f"{host}:{port}", method, path, headers, data, json, files)
+                self._send_request(
+                    socket, f"{host}:{port}", method, path, headers, data, json, files
+                )
             except OSError as exc:
                 last_exc = exc
                 ok = False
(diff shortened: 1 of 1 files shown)


bablokb:
Can someone check this please? It seems the tests are confused by the additional port information. The PR could be changed to only include the port if the port is not a standard port, but this would bloat the code without functional benefits.

justmobilize:
@bablokb you'll need to update all the tests. Random example: https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/82a656aa80797ae8c526b69061b9e75728d377c9/tests/reuse_test.py#L175-L182

If we sent the port on all of these, then we'll need to update all the tests.

I also don't think a single if statement for a non-standard port would add that much bloat

bablokb:
@justmobilize I changed the code: the hostname is passed as is (except to the socket-constructor), so the tests don't fail since they don't see the port anymore.

justmobilize:
@bablokb is it possible to add a test that hits that block?

bablokb:
For someone who knows how to write and verify tests, certainly. But this is currently beyond my reach. From what the code does, it would be sufficient to pass a hostname:port combination, e.g. www.adafruit.com:80 or www.adadruit.com:443. Although these are the default ports, they get passed since the port is provided explicitly.

